PR #17, opened 2016-12-14
include deployment information

sofuture:
We should include information on how to setup and run the selenium docker image here.


deploy/selenium.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
```
a-palchikov: I'd leave it chrome-driver specific (i.e. loose the 4444 port mapping)

PR #21, opened 2016-12-16
support more and configurable bandwagon fields

sofuture:
Add support for other bandwagon fields, if they exist. Also support configuring values for them.

### DIFF
--- a/e2e/model/ui/bandwagon.go
+++ b/e2e/model/ui/bandwagon.go
@@ -15,12 +15,12 @@ import (
 )
 
 type Bandwagon struct {
-	*framework.BandwagonConfig
+	framework.BandwagonConfig
 	page       *web.Page
 	domainName string
 }
 
-func OpenBandwagon(page *web.Page, domainName string, config *framework.BandwagonConfig) *Bandwagon {
+func OpenBandwagon(page *web.Page, domainName string, config framework.BandwagonConfig) Bandwagon {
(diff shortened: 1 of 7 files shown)


e2e/model/ui/bandwagon.go
```diff
@@ -15,13 +15,15 @@ import (
...
+	page         *web.Page
+	domainName   string
+	email        string
+	password     string
+	username     string
+	organization string
 }
 
-func OpenBandwagon(page *web.Page, domainName string, email string, password string) *Bandwagon {
+func OpenBandwagon(page *web.Page, domainName string, username string, password string, organization string, email string) *Bandwagon {
```
a-palchikov: I think it should take `BandwagonConfig` as parameter.

e2e/model/ui/bandwagon.go
```diff
@@ -35,7 +37,13 @@ func OpenBandwagon(page *web.Page, domainName string, email string, password str
 		BeFound(),
 		"should wait for bandwagon to load")
 
-	return &Bandwagon{page: page, email: email, password: password, domainName: domainName}
+	return &Bandwagon{page: page,
```
a-palchikov: `page` should slide on the next line

sofuture: how did `go fmt` let me get away with that? :confounded:

PR #43, opened 2017-04-07
Failed installation check

UnderGreen:


sofuture:
:shipit:

e2e/model/ui/installer/installer.go
```diff
@@ -130,10 +130,16 @@ func (i *Installer) StartInstallation() {
 }
 
 func (i *Installer) IsInstallCompleted() bool {
+	Expect(i.IsInstallFailed()).To(BeFalse())
```
alex-kovoy: _IsInstallCompleted_ is a helper method that returns true/fast but you are adding an assert (_expect()_) in it. Please create a separate method called something like MustBeCompleted()

PR #45, opened 2017-04-14
Sergei/endpoints fix

UnderGreen:


e2e/framework/framework.go
```diff
@@ -159,19 +158,14 @@ func InitializeCluster() {
...
-			// but tests like update need it
-			if Cluster != nil {
-				testState.EntryURL = strings.Replace(Cluster.OpsCenterURL(), "61009", "32009", -1)
-			}
-		}
 	} else {
 		Cluster, err = infra.New(config, TestContext.OpsCenterURL, provisioner)
 	}
 	Expect(err).NotTo(HaveOccurred())
+	testState.EntryURL = Cluster.OpsCenterURL()
```
UnderGreen: Here Cluster cannot be nil(because err will happen above), no need to check.

UnderGreen:
@sofuture, It will fix 2 issues:
1. Sometimes(for example for installation in wizard mode) installation can fail on the middle and `entry_url` will be nil in config.yaml.state. But we need it in a lot of places like crashreport downloading, backups, etc.
2. I fixed endpoints code, because robotest on last step go to founded endpoint[0] and try to open site page. But for anypoint we have 4 different types of endpoints and only one of it is site endpoint. I don't know how it worked before, because it should fail in 3 of 4 builds. ¯\_(ツ)_/¯

PR #46, opened 2017-04-18
Add a mode to only provision a cluster

UnderGreen:
Closes https://github.com/gravitational/gravity/issues/2036

### DIFF
--- a/e2e/framework/framework.go
+++ b/e2e/framework/framework.go
@@ -155,18 +155,11 @@ func InitializeCluster() {
 	}
 
 	var application *loc.Locator
-	switch {
-	case mode == provisionMode:
-		return
-	case mode == wizardMode:
-		{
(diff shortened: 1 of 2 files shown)


e2e/framework/framework.go
```diff
@@ -155,11 +155,18 @@ func InitializeCluster() {
...
 
 	var application *loc.Locator
-	if mode == wizardMode {
-		Cluster, application, err = infra.NewWizard(config, provisioner, installerNode)
-		TestContext.Application.Locator = application
-	} else {
-		Cluster, err = infra.New(config, TestContext.OpsCenterURL, provisioner)
+	switch {
+	case mode == provisionMode:
+		return
```
a-palchikov: If you exit early here, no cluster state is saved into a state file which means robotest binary will not be able to shut down the infrastructure. I think you don't need this as extra mode - just disabling the test should work w/o further modifications necessary.

PR #47, opened 2017-04-24
Docker All in One image

UnderGreen:


sofuture:
:shipit: once you add some documentation/example use.

UnderGreen:
@sofuture after `make build` robotest binary will be in **./build** directory. Have to change scripts here -> https://jenkins.gravitational.io/job/Robotest-publish/configure. Or use make targets, which I have added in this PR.

UnderGreen:
And, I think we should change Makefile for version/1.x brach for compatibility with our Jenkins job.

PR #49, opened 2017-04-28
Updated Makefile

UnderGreen:


sofuture:
I don't think we need to publish image to s3 and we can get rid of `e2e.test`, otherwise :shipit:

PR #57, opened 2017-05-10
Denis/azure/terraform

DenisMishin:
Initial set of changes to support multi-cloud provisioning

Breaking changes:
1. All Terraform parameters must come from config files, no more hard-coded params in .tf templates 
2. `install_url` must have schema - i.e. s3:// for S3 files


e2e/framework/test_context.go
```diff
@@ -149,11 +150,13 @@ type TestContextType struct {
 	// Wizard specifies whether wizard was used to bootstrap cluster
 	Wizard bool `json:"-" yaml:"-"`
 	// Provisioner defines the type of provisioner to use
-	Provisioner provisionerType `json:"provisioner" yaml:"provisioner" env:"ROBO_PROVISIONER"`
+	Provisioner provisionerType `json:"provisioner" yaml:"provisioner" `
+	// ProvisionTo defines cloud to deploy
+	ProvisionTo string `json:"provision_to" yaml:"provision_to" validate:"omitempty,eq=aws|eq=azure"`
```
r0mant: I would name it `CloudProvider` or something similar to be consistent with the terminology we use in the rest of the codebase (e.g. in gravity).

e2e/framework/test_context.go
```diff
@@ -149,11 +150,13 @@ type TestContextType struct {
...
 	// DumpCore specifies a command to collect all installation/operation logs
 	DumpCore bool `json:"-" yaml:"-"`
 	// StateDir specifies the location for test-specific temporary data
-	StateDir string `json:"state_dir" yaml:"state_dir" env:"ROBO_STATE_DIR"`
```
r0mant: Why are we getting rid of ability to override parameters via env vars btw?

e2e/framework/test_context.go
```diff
@@ -162,60 +165,63 @@ type TestContextType struct {
...
 	// Bandwagon defines the test configuration for post-install setup in bandwagon
 	Bandwagon BandwagonConfig `json:"bandwagon" yaml:"bandwagon"`
 	// WebDriverURL specifies optional WebDriver URL to use
-	WebDriverURL string `json:"web_driver_url,omitempty" yaml:"web_driver_url,omitempty" env:"ROBO_WEB_DRIVER_URL"`
+	WebDriverURL string `json:"web_driver_url,omitempty" yaml:"web_driver_url,omitempty" `
 	// Extensions groups arbitrary test step configuration
 	Extensions Extensions `json:"extensions,omitempty" yaml:"extensions,omitempty"`
 }
 
 type BandwagonConfig struct {
```
r0mant: Godocs on all exported structs/fields (below too).

e2e/framework/test_context.go
```diff
@@ -375,7 +351,22 @@ func newContextConfig(configFile string) error {
...
+	if err == nil {
+		return nil
+	}
+
+	log.Errorf("Configuration file %s has errors", configFile)
+	for _, fieldError := range err.(validator.ValidationErrors) {
+		log.Errorf("   Field %s=%v fails rule %s", fieldError.Field(), fieldError.Value(), fieldError.Tag())
+	}
+
+	return trace.Errorf("Configuration file fails validation test")
```
r0mant: Lower-case first letter of error messages, for better embedding: https://github.com/golang/go/wiki/CodeReviewComments#error-strings (above too).

infra/config.go
```diff
@@ -34,3 +34,63 @@ type StateNode struct {
 	// KeyPath defines the location of the SSH key
 	KeyPath string `json:"key_path"`
 }
+
+// AWSConfig describes AWS EC2 test configuration
+type AWSConfig struct {
+	AccessKey string `json:"access_key" yaml:"access_key" validate:"required"`
```
r0mant: Godocs on these two.

infra/config.go
```diff
@@ -34,3 +34,63 @@ type StateNode struct {
...
+// FIXME : replace with embedded validation rules
+func (r *AWSConfig) IsEmpty() bool {
+	return r.AccessKey == "" && r.SecretKey == ""
+}
+
+// AzureConfig specifies Azure cloud specific parameters
+// see https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-create-service-principal-portal
+// to obtain subscription_id, client_id, client_secret and tenant_id
+type AzureConfig struct {
+	SubscriptionId string `json:"subscription_id" yaml:"subscription_id" validate:"required"`
```
r0mant: Godocs.

infra/terraform/file_transfer.go
```diff
@@ -0,0 +1,78 @@
...
+    otherwise robotest completes but node is unsusable as there's still background process running
+
+ 2. waiting for boostrap script to complete (or maybe replacing its behind-the-scenes with more explicit preparation here)
+
+ 3. being sure which command exactly failed and avoiding monstrous && style pipelining
+
+ 4. support for SCP for local tarballs
+*/
+
+// TransferFile takes file URL which may be S3 or HTTP or local file and transfers it to the machine
```
r0mant: The comment seems to be misplaced.

infra/terraform/file_transfer.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	cmd := fmt.Sprintf(`echo Testing if bootstrap completed && test -f /var/lib/bootstrap_complete && \
+		echo Cleaning up && rm -rf %[1]s/installer/* 
+		echo Downloading installer %[5]s ... && %[2]s && \
+		echo Creating installer dir && mkdir -p %[1]s/installer && \
+		echo Unpacking installer && tar -xvf %[3]s -C %[1]s/installer && \
+		echo Launching command %[4]s && cd %[1]s/installer && %[4]s`, homeDir, fetchCmd, outFile, command, fileUrl)
+
+	// cmd = fmt.Sprintf(`echo Launching command %[4]s && cd %[1]s/installer && %[4]s`,
```
r0mant: Debugging remains?

infra/terraform/file_transfer.go
```diff
@@ -0,0 +1,78 @@
...
+
+	var fetchCmd string
+	var homeDir = fmt.Sprintf("/home/%s", t.sshUser)
+	var outFile string
+	if strings.HasSuffix(fileUrl, ".tar.gz") {
+		outFile = fmt.Sprintf("%s/installer.tar.gz", homeDir)
+	} else if strings.HasSuffix(fileUrl, ".tar") {
+		outFile = fmt.Sprintf("%s/installer.tar", homeDir)
+	} else {
+		return "", trace.Errorf("Unsupported installer packaging %s", fileUrl)
```
r0mant: That's pretty cryptic error message, something like "only .tar and .tar.gz are supported, got: .blah" would be clearer I think. Also, lowecase (as mentioned above).

infra/terraform/file_transfer.go
```diff
@@ -0,0 +1,78 @@
...
+// fileUrl - file to download, could be S3:// or http(s)://
+// command - to run out of installer package
+func (t *terraform) makeRemoteCommand(fileUrl, command string) (string, error) {
+	u, err := url.Parse(fileUrl)
+	if err != nil {
+		return "", trace.Wrap(err, "Parsing %s", fileUrl)
+	}
+
+	var fetchCmd string
+	var homeDir = fmt.Sprintf("/home/%s", t.sshUser)
```
r0mant: Not a big deal, but maybe choose a more reliable location for the file? Theoretically, user's home directory may not be under /home.

infra/terraform/file_transfer.go
```diff
@@ -0,0 +1,78 @@
...
+// TransferFile takes file URL which may be S3 or HTTP or local file and transfers it to the machine
+// fileUrl - file to download, could be S3:// or http(s)://
+// command - to run out of installer package
+func (t *terraform) makeRemoteCommand(fileUrl, command string) (string, error) {
+	u, err := url.Parse(fileUrl)
+	if err != nil {
+		return "", trace.Wrap(err, "Parsing %s", fileUrl)
+	}
+
+	var fetchCmd string
```
r0mant: I would declare this right before the switch where it's used. In general, I think it's a good idea to declare vars as close to their first usage location as possible.

infra/terraform/file_transfer.go
```diff
@@ -0,0 +1,78 @@
...
+		return "", trace.Errorf("Unsupported installer packaging %s", fileUrl)
+	}
+
+	switch u.Scheme {
+	case "s3":
+		if t.Config.AWS == nil {
+			return "", trace.Errorf("AWS config missing, cannot use S3 URLs %s", fileUrl)
+		}
+
+		fetchCmd = fmt.Sprintf(`AWS_ACCESS_KEY_ID=%s \
```
r0mant: Is this a reliable approach? Why aren't we using cloud init scripts for this as we used to?

infra/terraform/terraform.go
```diff
@@ -64,45 +71,73 @@ func NewFromState(config Config, stateConfig infra.ProvisionerState) (*terraform
...
+		if err == nil {
+			if withInstaller {
+				nodes := r.pool.Nodes()
+				if len(nodes) == 0 { // should not happen, and doesn't make sense to retry
+					return nil, trace.Errorf("Zero nodes were allocated")
+				}
+				r.installerIP = nodes[0].Addr()
+				return nodes[0], nil
+			}
+			return nil, nil
```
r0mant: Usually, if a function returns `nil` error, the called expects the first value to be usable, so I would avoid returning `nil, nil`. Return an appropriate error instead, like `trace.NotFound()` or `trace.BadParameter()`.

infra/terraform/terraform.go
```diff
@@ -64,45 +71,73 @@ func NewFromState(config Config, stateConfig infra.ProvisionerState) (*terraform
...
+		case <-ctx.Done():
+			return nil, trace.Wrap(err, "Terraform creation timed out")
+		case <-time.After(terraformRepeatAfter):
+			continue
+		}
+	}
+
+}
+
+func (r *terraform) terraform() (err error, retry bool) {
```
r0mant: That's very unusual to have error as a first return argument (unless it's the only one). Can you swap them?

lib/system/file.go
```diff
@@ -63,6 +64,72 @@ func CopyFileWithPerms(dst, src string, perm os.FileMode) error {
...
+	}
+
+	for _, entry := range entries {
+		srcPath := filepath.Join(src, entry.Name())
+		dstPath := filepath.Join(dst, entry.Name())
+
+		if entry.IsDir() {
+			err = copyAll(srcPath, dstPath, fileCount)
+			if err != nil {
+				return err
```
r0mant: `trace.Wrap(err)`

lib/system/file.go
```diff
@@ -63,6 +64,72 @@ func CopyFileWithPerms(dst, src string, perm os.FileMode) error {
...
+		}
+
+		if entry.Mode()&os.ModeSymlink != 0 {
+			log.Warning("Symlinks are not copied: %s", srcPath)
+			continue
+		}
+
+		err = CopyFile(srcPath, dstPath)
+		if err != nil {
+			return err
```
r0mant: `trace.Wrap(err)`

lib/system/file.go
```diff
@@ -63,6 +64,72 @@ func CopyFileWithPerms(dst, src string, perm os.FileMode) error {
...
+		*fileCount++
+		return CopyFile(src, filepath.Join(dst, filepath.Base(src)))
+	}
+
+	_, err = os.Stat(dst)
+	if err != nil && !os.IsNotExist(err) {
+		return trace.ConvertSystemError(err)
+	}
+
+	if err != nil && os.IsNotExist(err) {
```
r0mant: I think in this case it would be clearer to combine this with the above:

```go
if err != nil {
  if !os.IsNotExist(err) {
    return ...
  } 
  err = os.MkdirAll(...)
  ...
}
```

lib/system/file.go
```diff
@@ -63,6 +64,72 @@ func CopyFileWithPerms(dst, src string, perm os.FileMode) error {
...
+
+		if entry.IsDir() {
+			err = copyAll(srcPath, dstPath, fileCount)
+			if err != nil {
+				return err
+			}
+		}
+
+		if entry.Mode()&os.ModeSymlink != 0 {
+			log.Warning("Symlinks are not copied: %s", srcPath)
```
r0mant: Warningf?

e2e/framework/test_context.go
DenisMishin: https://docs.google.com/document/d/1lwbmjakDVan8haGwHUX65r5OR15tTSiFsz6qKKNVt5k/edit

just like we're getting rid of static variables; to allow concurrent multi-suite tests

e2e/framework/test_context.go
DenisMishin: that's pre-existing code; many structs meaning is unclear to me, and especially everything inside this test_context class is subject to massive changes in the coming days;

infra/config.go
DenisMishin: the meaning is described at the URL in the struct-level godoc.

infra/terraform/file_transfer.go
DenisMishin: that whole thing is temporary implementation, see FIXME at top

infra/terraform/file_transfer.go
DenisMishin: as per aws cli docs; also it's not necessarily an AWS instance, and in general I don't like global inits which is all over robotest code. 

once that'll be replaced with more fine-grained SSH, it'll help to shield sensitive key stuff from logs.

DenisMishin: also cloud init need be kept at bare minimum, no test logic like fetching installers should live there.

lib/system/file.go
```diff
@@ -63,6 +64,72 @@ func CopyFileWithPerms(dst, src string, perm os.FileMode) error {
 	return nil
 }
 
+// Recursively copy
+// dst must always be a directory
+// src may be either a dir or a file
+func CopyAll(src, dst string) (err error, fileCount uint) {
```
a-palchikov: Please return errors last

lib/system/file.go
```diff
@@ -63,6 +64,72 @@ func CopyFileWithPerms(dst, src string, perm os.FileMode) error {
...
+	fileCount = 0
+	err = copyAll(src, dst, &fileCount)
+	return err, fileCount
+}
+
+func copyAll(src, dst string, fileCount *uint) (err error) {
+	src = filepath.Clean(src)
+	dst = filepath.Clean(dst)
+
+	log.Debugf("CopyAll %s %s", src, dst)
```
a-palchikov: `%s -> %s` to define the operation direction

lib/system/file.go
```diff
@@ -63,6 +64,72 @@ func CopyFileWithPerms(dst, src string, perm os.FileMode) error {
...
+
+	for _, entry := range entries {
+		srcPath := filepath.Join(src, entry.Name())
+		dstPath := filepath.Join(dst, entry.Name())
+
+		if entry.IsDir() {
+			err = copyAll(srcPath, dstPath, fileCount)
+			if err != nil {
+				return err
+			}
```
a-palchikov: You need to `continue` here - otherwise the function will try to `CopyFile` a directory.
On a side note - why reinvent the [`filepath.Walk`](https://golang.org/pkg/path/filepath/#Walk)?

assets/terraform/aws/bootstrap.sh
```diff
@@ -0,0 +1,36 @@
...
+mount /var/lib/gravity
+mount /var/lib/data
+mkdir -p /var/lib/gravity/planet/etcd
+mount /var/lib/gravity/planet/etcd
+
+chown -R 1000:1000 /var/lib/gravity /var/lib/data /var/lib/gravity/planet/etcd
+sed -i.bak 's/Defaults    requiretty/#Defaults    requiretty/g' /etc/sudoers
+
+# robotest might SSH before bootstrap script is complete (and will fail)
+touch /var/lib/bootstrap_complete
```
DenisMishin: its just tested for existence

infra/terraform/file_transfer.go
```diff
@@ -0,0 +1,78 @@
...
+	case "https":
+		fetchCmd = fmt.Sprintf("wget %s -O %s/", fileUrl, outFile)
+	case "":
+	case "gs":
+	default:
+		// TODO : implement SCP and GCLOUD methods
+		return "", fmt.Errorf("Unsupported URL schema %s", fileUrl)
+	}
+
+	cmd := fmt.Sprintf(`echo Testing if bootstrap completed && test -f /var/lib/bootstrap_complete && \
```
DenisMishin: that's temporary implementation, to be replaced.

sofuture:
What does 1. mean? Can't use TF defaults?

assets/terraform/aws/bootstrap.sh
```diff
@@ -0,0 +1,36 @@
...
+apt install -y python-pip lvm2
+pip install --upgrade awscli
+
+set -euo pipefail
+
+pvcreate -ff /dev/xvde
+vgcreate data /dev/xvde
+lvcreate -n var_lib_gravity -l 20%VG data
+lvcreate -n var_lib_data -l 20%VG data
+lvcreate -n etcd -l 60%VG data
```
sofuture: In practice this needs a real 'physical' device of it's own. Instance storage if the instance has it, or a 50GB 2500IOPS io1 SSD on AWS.

DenisMishin: etcd or each ?

DenisMishin: >> Can't use TF defaults?
Yes, TF would be just provisioning templates; all configuration values would come top-down.

assets/terraform/aws/bootstrap.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
sofuture: Once we are done bootstrapping, we should disable internet egress with iptables. We've shipped regressions of things that needing internet before, without doing that. If we find exceptions that we want to allow in CI, like say NTP, can enable those port by port.

See: https://github.com/gravitational/anypoint/blob/develop/terraform/onprem.tf#L454

assets/terraform/aws/network.tf
```diff
@@ -0,0 +1,44 @@
...
+
+    # install wizard
+    ingress {
+        from_port   = 61009
+        to_port     = 61009
+        protocol    = "tcp"
+        cidr_blocks = ["0.0.0.0/0"]
+    }
+
+    ingress {
```
sofuture: Out of ultra caution, we should not allow blanket intra-node communcation -- and should make it port-by-port. Luckily I have a relatively complete example of this :)

```
    # installer ports
    ingress {
        from_port = 61008
        to_port = 61010
        protocol = "tcp"
        self = true
    }

    # installer ports
    ingress {
        from_port = 61022
        to_port = 61024
        protocol = "tcp"
        self = true
    }

    # bandwidth checker (for pre-checks)
    ingress {
        from_port   = 4242
        to_port     = 4242
        protocol    = "tcp"
        cidr_blocks = ["0.0.0.0/0"]
    }

    # k8s health check
    ingress {
        from_port = 10250
        to_port = 10250
        protocol = "tcp"
        cidr_blocks = ["0.0.0.0/0"]
    }

    # serf peer-to-peer
    ingress {
        from_port = 7496
        to_port = 7496
        protocol = "tcp"
        self = true
    }

    # kubernetes apiserver insecure
    ingress {
        from_port = 8080
        to_port = 8080
        protocol = "tcp"
        self = true
    }

    # kubernetes apiserver secure
    ingress {
        from_port = 6443
        to_port = 6443
        protocol = "tcp"
        self = true
    }

    # etcd peer-to-peer
    ingress {
        from_port = 2380
        to_port = 2380
        protocol = "tcp"
        self = true
    }

    # etcd client
    ingress {
        from_port = 2379
        to_port = 2379
        protocol = "tcp"
        self = true
    }

    # etcd legacy client
    ingress {
        from_port = 4001
        to_port = 4001
        protocol = "tcp"
        self = true
    }

    # etcd legacy peer-to-peer
    ingress {
        from_port = 7001
        to_port = 7001
        protocol = "tcp"
        self = true
    }

    # docker registry
    ingress {
        from_port = 5000
        to_port = 5000
        protocol = "tcp"
        self = true
    }

    # overlay networking
    ingress {
        from_port = 8472
        to_port = 8472
        protocol = "udp"
        self = true
    }

    # gravity profiling port
    ingress {
        from_port = 6060
        to_port = 6060
        protocol = "tcp"
        self = true
    }

    # teleport services (SSH server, proxy service, proxy tunnel, auth service)
    ingress {
        from_port = 3022
        to_port = 3025
        protocol = "tcp"
        self = true
    }

    # teleport web
    ingress {
        from_port = 3080
        to_port = 3080
        protocol = "tcp"
        self = true
    }

    # internal k8s services
    ingress {
        from_port = 30000
        to_port = 32767
        protocol = "tcp"
        self = true
    }

    # kubernetes kubelet
    ingress {
        from_port = 10248
        to_port = 10249
        protocol = "tcp"
        self = true
    }

    # kubernetes kubelet
    ingress {
        from_port = 10255
        to_port = 10255
        protocol = "tcp"
        self = true
    }

    # gravity services (ops service, pack service, health test)
    ingress {
        from_port = 3008
        to_port = 3010
        protocol = "tcp"
        self = true
    }

    # planet agent RPC
    ingress {
        from_port = 7575
        to_port = 7575
        protocol = "tcp"
        self = true
    }

    # serf RPC
    ingress {
        from_port = 7373
        to_port = 7373
        protocol = "tcp"
        self = true
    }
```

assets/terraform/aws/node.tf
```diff
@@ -0,0 +1,43 @@
...
+    root_block_device {
+        delete_on_termination = true
+        volume_type = "io1"
+        volume_size = "50"
+        iops = 500
+    }
+
+    # etc and other data device
+    ebs_block_device = {
+        volume_size = "500"
```
sofuture: We can make this way smaller: 100gb standard SSD should be fine (split etcd into it's own 50gb 2500IOPS io1 though)

assets/terraform/aws/node.tf
```diff
@@ -0,0 +1,43 @@
...
+        volume_type = "io1"
+        device_name = "/dev/xvde"
+        iops = 3000
+        delete_on_termination = true
+    }
+
+    # gravity/docker data device
+    ebs_block_device = {
```
sofuture: This is unused in bootstrap, I assume it's a device for docker device mapper. It can be made much smaller and lower performance as well: 100GB, 2000IOPS is plenty

assets/terraform/azure/bootstrap.sh
```diff
@@ -0,0 +1,36 @@
...
+apt install -y python-pip lvm2
+pip install --upgrade awscli
+
+set -euo pipefail
+
+pvcreate -ff /dev/sdc
+vgcreate data /dev/sdc
+lvcreate -n var_lib_gravity -l 20%VG data
+lvcreate -n var_lib_data -l 20%VG data
+lvcreate -n etcd -l 30%VG data
```
sofuture: same issue with etcd -- I know disks on azure are pretty different (worse), so maybe doesn't apply, but I would assume it does.

assets/terraform/azure/bootstrap.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
sofuture: ditto with egress

assets/terraform/azure/network.tf
```diff
@@ -0,0 +1,68 @@
...
+    priority                   = 1020
+    direction                  = "Inbound"
+    access                     = "Allow"
+    protocol                   = "TCP"
+    source_port_range          = "*"
+    destination_port_range     = 61009
+    source_address_prefix      = "*"
+    destination_address_prefix = "*"
+  }
+}
```
sofuture: ditto with whitelisting intra-cluster ports

e2e/framework/test_context.go
```diff
@@ -375,7 +351,22 @@ func newContextConfig(configFile string) error {
...
+	}
+
+	err = validator.New().Struct(TestContext)
+	if err == nil {
+		return nil
+	}
+
+	log.Errorf("Configuration file %s has errors", configFile)
+	for _, fieldError := range err.(validator.ValidationErrors) {
+		log.Errorf("   Field %s=%v fails rule %s", fieldError.Field(), fieldError.Value(), fieldError.Tag())
```
sofuture: Thank you for this :+1:

infra/config.go
sofuture: Still, it makes for nicer autogenerated docs to follow the idiom of 'FooBar ...' comments on every exported `FooBar`

infra/terraform/file_transfer.go
sofuture: I agree, fetching installers should be separate from provisioning infrastructure -- it's a wart in how customer uses this right now, they have a `while aws sleep` loop in cloud-init to get installer. This starts to give us a little more control over flow.

assets/terraform/aws/bootstrap.sh
sofuture: Just etcd -- and updated figures: 30gb 1500IOPS.

e2e/framework/framework.go
```diff
@@ -298,7 +298,7 @@ func fetchOpsCenterLogs() error {
 		return trace.Wrap(err, "failed to collect site report")
 	}
 
-	err = system.CopyFile(output, tempOutput.Name())
+	err = system.CopyFile(tempOutput.Name(), output)
```
DenisMishin: stdlib is counter intuitive by itself... when you do cp its always source, then destination

assets/terraform/aws/bootstrap.sh
DenisMishin: I'll make all TF template enhancements part of a separate PR after I test all deployment params such as different distros, etc; don't hold this specific PR for too long.

infra/terraform/terraform.go
```diff
@@ -64,45 +71,73 @@ func NewFromState(config Config, stateConfig infra.ProvisionerState) (*terraform
 }
 
 func (r *terraform) Create(withInstaller bool) (installer infra.Node, err error) {
-	err = system.CopyFile(filepath.Join(r.stateDir, "terraform.tf"), r.ScriptPath)
-	if err != nil {
-		return nil, trace.Wrap(err)
-	}
+	ctx, cancel := context.WithTimeout(context.Background(), terraformTimeout)
```
DenisMishin: that would be a chain reaction; it'll have to go top-down between most functions in infra, framework and lib package; which is fine but it should be subject to separate PR, likely coupled with changes to more fine-grained SSH command execution. 

please approve as-is.

a-palchikov: Ok

PR #64, opened 2017-05-19
Tests/functional

DenisMishin:
CLI based tests and infra


infra/infra.go
```diff
@@ -126,11 +129,16 @@ type NodePool interface {
...
 // Node defines an interface to a remote node
 type Node interface {
-	// Addr returns the address of the node
+	// Addr returns public address of the node
 	Addr() string
+	// PrivateAddr returns the private address of the node
+	PrivateAddr() string
 	// Connect connects to this node and returns a new session object
 	// that can be used to execute remote commands
 	Connect() (*ssh.Session, error)
```
a-palchikov: We don't need `Connect` as part of `Node` anymore - one can always create a `Session` from a `Client`. Besides, the implementation is identical and hence duplicate in each provisioner.

I think internal unit-tests are out-of-date w.r.t changes to this interface.

lib/ssh/commands.go
```diff
@@ -0,0 +1,32 @@
...
+// Cmd defines command to execute on remote host
+type Cmd struct {
+	// Command is what is passed to remote shell
+	Command string
+	// Env are any environment variables to be set
+	// Note that environment vars must be explicitly enabled in /etc/ssh/sshd.conf `AcceptEnv` directive
+	Env map[string]string
+}
+
+// RunCommands execute commands sequentially
```
a-palchikov: s/execute/executes as the function name is treated as just a name

lib/ssh/run_parse.go
```diff
@@ -0,0 +1,139 @@
...
+	if exit != 0 {
+		return trace.Errorf("%s returned %d", cmd, exit)
+	}
+
+	return nil
+}
+
+// RunAndParse runs remote SSH command with environment variables set by `env`
+// exitStatus is -1 if undefined
+func RunAndParse(ctx context.Context, logFn LogFnType, client *ssh.Client, cmd string, env map[string]string, parse OutputParseFn) (out interface{}, exitStatus int, err error) {
```
a-palchikov: The function expects a `ssh.Session`, and the logging detail can be setup prior to the call:
```go
session, err := client.NewSession()
if err != nil {
	return trace.Wrap(err, "failed to create a session")
}
entry := log.WithFields(log.Fields{"client": client.RemoteAddr()})
exitStatus, err := RunAndParse(ctx, session, cmd, env, parseFn, entry)

func RunAndParse(ctx context.Context, session *ssh.Session, cmd string, env map[string]string, parse OutputParseFn, entry *log.Entry) (out interface{}, exitStatus int, err error) {
	// ...
	entry.Debugf("stderr: %s", out)
}
```
This way you don't need to spill too much detail into the function itself.

And the function itself feels too heavy.

lib/ssh/run_parse.go
```diff
@@ -0,0 +1,139 @@
...
+		for {
+			line, err := r.ReadString('\n')
+			if err != nil {
+				return
+			}
+			logFn("[%v] %s [stderr] %s", client.RemoteAddr(), cmd, line)
+		}
+	}()
+
+	outCh := make(chan interface{}, 2)
```
a-palchikov:
```go
var outCh chan resp

type resp struct {
	out []byte
	err error
}
```
why the buffer of 2? There's just a single producer

lib/ssh/run_parse.go
```diff
@@ -0,0 +1,139 @@
...
+	go func() {
+		logFn("[%v] (starting) %s", client.RemoteAddr(), cmd)
+		runCh <- session.Run(fmt.Sprintf("%s %s", strings.Join(envStrings, " "), cmd))
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil, exitStatusUndefined,
+			trace.Errorf("[%v] %s timed out, sending SIGTERM: %v",
+				client.RemoteAddr(), cmd, session.Signal(ssh.SIGTERM))
```
a-palchikov: can you extract the `Signal` call (so it's not part of `Errorf` call)?

lib/ssh/ssh_test.go
```diff
@@ -0,0 +1,105 @@
...
+		client,
+		"sleep 100",
+		nil,
+		ParseDiscard)
+	assert.Error(t, err)
+	assert.Equal(t, exitStatusUndefined, exit)
+}
+
+func testExitErr(t *testing.T, client *ssh.Client) {
+	_, exit, err := RunAndParse(context.Background(), t.Logf, client, "ls /nosuchdir", nil, ParseDiscard)
```
a-palchikov: better use `false` as an always failing command

suite/install.go
```diff
@@ -0,0 +1,52 @@
...
+		go func(n gravity.Gravity) {
+			// TODO: how to properly define node role ?
+			errs <- n.Join(ctx, gravity.JoinCmd{
+				PeerAddr: master.Node().PrivateAddr(),
+				Token:    token,
+				Role:     defaultRole})
+		}(node)
+	}
+
+	for range nodes {
```
a-palchikov: Handle nodes never returning from install/join call with a timeout?

infra/gravity/gravity_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	sshutils "github.com/gravitational/robotest/lib/ssh"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGravityOutput(t *testing.T) {
+	for _, s := range []struct {
+		label string
+		fn    sshutils.OutputParseFn
```
DenisMishin: as discussed will be changed after implemented in gravity repo in next few days.

infra/gravity/log.go
```diff
@@ -0,0 +1,20 @@
...
+func init() {
+	log.SetFlags(0)
+}
+
+func Logf(t *testing.T, prefix string) sshutils.LogFnType {
+	return func(format string, args ...interface{}) {
+		t.Logf(format, args...)
+		log.Printf(fmt.Sprintf("[%s] %s", prefix, format), args...)
+	}
+}
```
DenisMishin: as discussed, will stay as is: in non-verbose mode, will just be equal to t.Logf and thus will log line numbers correctly.

infra/gravity/provision.go
```diff
@@ -0,0 +1,257 @@
...
+	dockerDevice string
+	user         string
+	os           string
+	installDir   string
+	installerUrl string
+	tf           terraform.Config
+	env          map[string]string
+}
+
+// makeDynamicParams takes base config, validates it and returns TerraformConfig
```
DenisMishin: it's not related to just terraform, so naming is right.

infra/terraform/terraform.go
```diff
@@ -250,6 +249,7 @@ func (r *terraform) boot() (output string, err error) {
 	out, err := r.command([]string{
 		"apply",
 		"-var", fmt.Sprintf("nodes=%d", r.NumNodes),
+		"-var", fmt.Sprintf("os=%s", r.OS),
```
DenisMishin: as discussed, TF destroy is not very reliable method of tearing down, and it'll be external service which would ensure cleanup by timeout. anyway right now destory is a no-op, and it'll be addressed in next few days.

infra/infra.go
DenisMishin: for legacy code compatibility

suite/install.go
DenisMishin: handled by ctx

infra/gravity/provision.go
a-palchikov: I meant that that the original comment states it returns a TerraformConfig while it does not:
`// makeDynamicParams takes base config, validates it and returns TerraformConfig`

Makefile
```diff
@@ -1,82 +1,62 @@
...
+docker-images:
+	cd docker && $(MAKE) publish
 
 .PHONY: publish-binary-into-s3
 publish-binary-into-s3:
-ifeq (, $(shell which aws))
-	$(error "No aws command in $(PATH)")
-endif
-	aws $(S3_OPTS) s3 cp ./build/robotest $(BUILD_BUCKET_URL)/robotest
+	ifeq (, $(shell which aws))
```
UnderGreen: It's not a nitpick. Without normal indentation `ifeq` will not work. I know, because I've found this issue several weeks ago.

PR #66, opened 2017-05-23
Denis/install options

DenisMishin:
1. more shared utils - SCP, remote scripts, sync utils, etc.
2. gravity cluster manipulation
3. rollback to terraform 0.9.3
4. code reorg : better separation of tests from common operations helper


infra/gravity/config.go
```diff
@@ -0,0 +1,123 @@
...
+func (config *ProvisionerConfig) WithTag(tag string) *ProvisionerConfig {
+	cfg := *config
+	cfg.tag = fmt.Sprintf("%s-%s", cfg.tag, tag)
+	cfg.stateDir = filepath.Join(cfg.stateDir, tag)
+
+	return &cfg
+}
+
+// WithNodes returns copy of config with specific number of nodes
+func (config *ProvisionerConfig) WithNodes(nodes uint) *ProvisionerConfig {
```
a-palchikov: Like here, no need for both pointer-receiver and pointer result

infra/gravity/provision.go
```diff
@@ -210,7 +164,27 @@ func Provision(ctx context.Context, t *testing.T, baseConfig ProvisionerConfig)
...
+
+	return sorted(gravityNodes), p.Destroy, nil
+}
+
+// sort Interface implementation
+type gravityList []Gravity
+
+func (g gravityList) Len() int { return len(g) }
+func (g gravityList) Swap(i, j int) {
+	tmp := g[i]
```
a-palchikov:
```go
g[i], g[j] = g[j], g[i]
```

infra/gravity/provision.go
```diff
@@ -210,7 +164,27 @@ func Provision(ctx context.Context, t *testing.T, baseConfig ProvisionerConfig)
...
 	}
 
-	return gravityNodes, p.Destroy, nil
+	logFn("[complete] OS=%s NODES=%d TAG=%s DIR=%s", baseConfig.os, baseConfig.nodeCount, baseConfig.tag, baseConfig.stateDir)
+
+	return sorted(gravityNodes), p.Destroy, nil
+}
+
+// sort Interface implementation
+type gravityList []Gravity
```
a-palchikov: s/gravityList/byPrivateAddr - to make the sort order clear

infra/gravity/runtest.go
```diff
@@ -0,0 +1,43 @@
...
+	Sequential = false
+)
+
+var reFuncName = regexp.MustCompile(`^([\w\/\.]+)\/([\d\w]+)\.([\d\w]+)$`)
+
+// Run is a handy wrapper around test runner to have a properly assigned tag, state dir, etc.
+func Run(ctx context.Context, t *testing.T, config *ProvisionerConfig, fn TestFunc, payload interface{}, parallel bool) bool {
+	fnName := runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
+	vars := reFuncName.FindStringSubmatch(fnName)
+	require.Len(t, vars, 4, "internal error")
```
a-palchikov: not a helpful message

lib/ssh/run_parse.go
```diff
@@ -60,34 +61,45 @@ func RunAndParse(ctx context.Context, logFn LogFnType, client *ssh.Client, cmd s
 		return nil, exitStatusUndefined, trace.Wrap(err)
 	}
 
+	outCh := make(chan interface{}, 1)
```
a-palchikov: no reason for interface{}, just pack into a struct {error; []byte}

lib/ssh/scp.go
```diff
@@ -0,0 +1,88 @@
...
+func PutFile(ctx context.Context, logFn utils.LogFnType, client *ssh.Client, srcPath, dstDir string) (remotePath string, err error) {
+	session, err := client.NewSession()
+	if err != nil {
+		return "", trace.Wrap(err)
+	}
+	defer session.Close()
+
+	f, err := os.Open(srcPath)
+	if err != nil {
+		return "", trace.Wrap(err)
```
a-palchikov: There's `trace.ConvertSystemError` from os-specific errors to trace errors (if necessary)

lib/ssh/scp.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	errCh := make(chan error, 2)
+	go scpSendFile(session, f, fi, errCh)
+
+	go func() {
+		err := session.Run(fmt.Sprintf("/usr/bin/scp -tr %s", dstDir))
+		errCh <- err
```
a-palchikov: this error will not have any trace. I guess it will be harder to find the problem

lib/ssh/scp.go
```diff
@@ -0,0 +1,88 @@
...
+		return remotePath, nil
+	}
+}
+
+func scpSendFile(session *ssh.Session, file *os.File, fi os.FileInfo, errCh chan error) {
+	cmd := fmt.Sprintf("C%04o %d %s\n", fi.Mode()&os.ModePerm, fi.Size(), fi.Name())
+
+	out, err := session.StdinPipe()
+	if err != nil {
+		errCh <- trace.Wrap(err)
```
a-palchikov: this does not look super convenient. Split the work between another function to do all the work and return an error, then call it and send the result to the channel

lib/utils/collect_errors.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+
+	"github.com/gravitational/trace"
+)
+
+// WaitForErrors collects errors from channel provided, honouring timeout
+func CollectErrors(ctx context.Context, count int, errChan chan error) error {
+	if count <= 0 {
```
a-palchikov: how about just using `cap(errCh)`?

infra/gravity/config.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	if cfg.stateDir = stateDir; cfg.stateDir == "" {
+		cfg.stateDir, err = ioutil.TempDir("", "robotest")
+		require.NoError(t, err, "tmp dir")
+	}
+
+	cfg.stateDir = filepath.Join(cfg.stateDir, cfg.tag)
+
+	return &cfg
```
DenisMishin: yep, I had some functions which returned it together with error, so although functions are gone, the pointer remains in case the need will arise again.

infra/gravity/runtest.go
```diff
@@ -0,0 +1,43 @@
...
+
+const (
+	Parallel   = true
+	Sequential = false
+)
+
+var reFuncName = regexp.MustCompile(`^([\w\/\.]+)\/([\d\w]+)\.([\d\w]+)$`)
+
+// Run is a handy wrapper around test runner to have a properly assigned tag, state dir, etc.
+func Run(ctx context.Context, t *testing.T, config *ProvisionerConfig, fn TestFunc, payload interface{}, parallel bool) bool {
```
a-palchikov: I'm not a big fan of `interface{}`. In this case, I'd encapsulate the payload in the `TestFunc` - e.g. instead of relaying it starting in Run, just create a TestFunc as a closure with the necessary parameters.

DenisMishin: it's used inside loops, an encapsulating payload there results in less readable code compared to current:

```
gravity.Run(ctx, t, config, cycleInstall, cycleInstallParam{nodes: size, cycles: 15}, gravity.Parallel)
```

a-palchikov: How's this less readable?
```go
gravity.Run(ctx, t, config, cycleInstall(cycleInstallParam{nodes: size, cycles: 15}), gravity.Parallel)

func cycleInstall(params cycleInstallParams) {
	return func(ctx context.Context, t *testing.T, baseConfig *gravity.ProvisionerConfig) {
		cfg := baseConfig.WithNodes(param.nodes)
		nodes, destroyFn, err := gravity.Provision(ctx, t, cfg)
		require.NoError(t, err, "provision nodes")
		defer gravity.Destroy(t, destroyFn)

		var c uint
		for c = 1; c <= param.cycles; c++ {
			gravity.OfflineInstall(ctx, t, nodes)
			gravity.Uninstall(ctx, t, nodes)
		}
	}
}
```

DenisMishin: closures have their set of issues with variable scoping
I changed into closures but IMHO they bring more bugs then interface{} usage

lib/ssh/commands.go
```diff
@@ -20,14 +16,14 @@ type Cmd struct {
 }
 
 // RunCommands executes commands sequentially
-func RunCommands(ctx context.Context, logFn utils.LogFnType, client *ssh.Client, commands []Cmd) error {
+func RunCommands(ctx context.Context, node SshNode, commands []Cmd) error {
```
a-palchikov: minor: the acronyms are usually not camel-cased (Ssh vs SSH): https://github.com/golang/go/wiki/CodeReviewComments#initialisms

suite/suite_test.go
```diff
@@ -32,18 +35,37 @@ func TestMain(t *testing.T) {
 		t.Fatal("options required")
 	}
 
-	testFn, there := suites[*testSuite]
+	config := gravity.LoadConfig(t, *configFile, *stateDir, *tag)
+
+	osFlavors := []string{}
+	for _, flavor := range strings.Split(*osFlag, ",") {
```
a-palchikov: shorter
```go
osFlavors := strings.Split(*osFlag, ",")
```
Off-topic: define a type of comma-separated string flags if more list parameters are to come:

```go
var osFlavors valueList
flag.Var(&osFlavors, "os", "ubuntu", "comma delimited list of OSes")

type valueList []string
func (r *valueList) String() string {return strings.Join(r, ",")}
func (r *valueList) Set(value string) error {
	*r = strings.Split(value, ",")
	return nil
}
```

PR #68, opened 2017-05-25
Alexey/e2e/refactoring

alex-kovoy:
- contains refactoring of the e2e/specs and e2e/ui packages


e2e/framework/test_context.go
```diff
@@ -209,10 +208,11 @@ type TestContextType struct {
...
 
 type BandwagonConfig struct {
-	Organization string `json:"organization" yaml:"organization" `
-	Username     string `json:"username" yaml:"username" `
-	Password     string `json:"password" yaml:"password" `
-	Email        string `json:"email" yaml:"email" `
+	Organization string `json:"organization" yaml:"organization" env:"ROBO_BANDWAGON_ORGANIZATION"`
+	Username     string `json:"username" yaml:"username" env:"ROBO_BANDWAGON_USERNAME"`
+	Password     string `json:"password" yaml:"password" env:"ROBO_BANDWAGON_PASSWORD"`
+	Email        string `json:"email" yaml:"email" env:"ROBO_BANDWAGON_EMAIL"`
```
DenisMishin: 1. I thought we agreed on eliminating global initializers such as ENV variables. Also you don't need them here. 
2. You don't use JSON tag
3. This config is part of Bandwagon defined as 
```
	// Bandwagon specifies bandwagon creation details
	Bandwagon *BandwagonConfig `json:"bandwagon,omitempty"`
```
- please add `validate` tag to the fields themselves : they either need be filled in in full OR replaced with default one as a whole if it came as `nil` from config file  - don't cherry pick default values as in `func (b *Bandwagon) SubmitForm(config framework.BandwagonConfig)`

e2e/uimodel/ui.go
```diff
@@ -0,0 +1,45 @@
...
+	return opscenter.Open(u.page, URL)
+}
+
+// GoToBandwagon navigates to bandwagon page and returns bandwagon object
+func (u *UI) GoToBandwagon(domainName string) bandwagon.Bandwagon {
+	return bandwagon.Open(u.page, domainName)
+}
+
+// EnsureUser navigates to given URL and ensures that a user is logged in
+func (u *UI) EnsureUser(URL string) {
```
DenisMishin: I think it'll be more readable if it would be two functions 
```
func (u *UI) Go(url string) (*UI)
func (u UI) RequireUserLoggedIn()
```

e2e/onprem_test.go
```diff
@@ -1,13 +1,127 @@
...
+	ctx := framework.TestContext
+	var ui uimodel.UI
+	var domainName string
+
+	BeforeEach(func() {
+		domainName = ctx.ClusterName
+		ui = uimodel.Init(f.Page)
+	})
+
+	It("should provision a new cluster [provisioner:onprem_install]", func() {
```
DenisMishin: I thought we discussed splitting that into dedicated functions so that tests could be composed out of them

e2e/uimodel/agent/agent.go
```diff
@@ -3,20 +3,23 @@ package agent
...
+	// Hostname is agent server hostname
 	Hostname    string
 	cssSelector string
 	page        *web.Page
 }
 
+// CreateAgentServer finds agent servers on the page and creates ui models
 func CreateAgentServer(page *web.Page, cssSelector string) AgentServer {
 	elem := page.Find(cssSelector)
 	Expect(elem).To(BeFound())
```
DenisMishin: As I understand it'll panic with very minimal messaging about what happened if the element is not there. At least maybe make is say the URL of page and CSS element name.

e2e/uimodel/utils/utils.go
```diff
@@ -135,8 +161,7 @@ func PauseForServerListRefresh() {
 }
 
 func Pause(params ...time.Duration) {
```
DenisMishin: doesn't seem to be used anywhere

e2e/uimodel/user/login.go
```diff
@@ -119,12 +73,37 @@ func (u *User) LoginWithGoogle() {
...
 }
+
+// EnsureUser navigates to given URL and ensures that a user is logged in
+func EnsureUser(page *web.Page, URL string) {
+	log.Infof("navigating to %q", URL)
+	login := framework.TestContext.Login
+	Expect(page.Navigate(URL)).To(Succeed())
+	count, _ := page.FindByClass("grv-user-login").Count()
+
+	if count != 0 {
```
DenisMishin: I wonder if it's good approach to try to login behind the scenes on every step, given how ofter that function is used. 
User should be logged in in the beginning of a test, this function should be called `Navigate()` or `Go()` and another function `RequiredLoggedIn()` could assert that user is still logged in.

e2e/uimodel/bandwagon/bandwagon.go
```diff
@@ -0,0 +1,88 @@
...
+	Expect(page.Navigate(url)).To(Succeed(), "should open bandwagon")
+	Eventually(page.FindByClass("my-page-btn-submit"), defaults.FindTimeout).
+		Should(BeFound(), "should wait for bandwagon to load")
+	return Bandwagon{page}
+}
+
+// SubmitForm submits bandwagon form
+func (b *Bandwagon) SubmitForm(config framework.BandwagonConfig) {
+	log.Infof("trying to submit bandwagon form")
+	if config.Organization == "" {
```
DenisMishin: not sure if it's a good approach to fill-in just some of the values with defaults. 
if no config is provided (config as a pointer == nil) then take a default one.

e2e/framework/test_context.go
alex-kovoy: it was a bad merge, will remove

e2e/uimodel/ui.go
alex-kovoy: Not sure how it would make it more readable when the goal is to ensure that user is logged in. We have several cases like one-time installer link or wizard mode when it's unclear upfront if login is required.

e2e/onprem_test.go
alex-kovoy: a lot of code went to the models providing a high level abstractions which are supposed to be used on the actual tests.

if we need to call 'expand with XXX' in _other_ tests then we can refactor above *It* into the method.

e2e/uimodel/agent/agent.go
alex-kovoy: i will be refining error messages as I proceed with other tasks. it would've been too much changes if i tried to address all in one PR.

e2e/uimodel/ui.go
DenisMishin: 1. The naming is obscure right now: I had to look inside the source of a function to understand what is it actually doing. 
2. Then I have an issue with this concept of logging in behind the scenes every time it's invoked. This approach would not allow you to catch i.e. a bug when some step user gets kicked out. Also you have this `BeforeEach()` where you could put an assert that user is logged in part of the test steps which expect user to be logged in already.
3. When you have a case as you described, when you're not sure, why not just check it and explicitly log in if needed.

e2e/uimodel/user/login.go
alex-kovoy: > behind the scenes on every step, given how ofter that function is used.

Not sure I understand. We ensure that the user is logged in before each 'test' since they can be called separately. Like you run an install tests and later you run expand tests against existing state.

e2e/uimodel/bandwagon/bandwagon.go
alex-kovoy: yes, this part will be revisited in the upcoming PRs.

e2e/uimodel/agent/agent.go
DenisMishin: I think this just needs some top-level handler to dump web browser current URL along with screenshot. You do see line number; what you don't know is why the selector is not there.

e2e/uimodel/site/siteservers.go
```diff
@@ -0,0 +1,307 @@
...
+	utils.SelectRadio(p.site.page, ".grv-control-radio", func(value string) bool {
+		return strings.HasPrefix(value, profileLabel)
+	})
+}
+
+func (p *ServerPage) getProfileLabel(profileName string) string {
+	var profileLabel string
+	siteName := p.site.domainName
+	const jsTemplate = `
+		var profileName = "%v";
```
a-palchikov: I'd actually like this to be a text/template - this is hard to see where which variables go.

e2e/uimodel/site/siteservers.go
```diff
@@ -0,0 +1,307 @@
...
+		var server = null;		
+		var nodeProfilers = reactor.evaluate(["sites", "%v", "app", "manifest", "nodeProfiles"])
```
a-palchikov: isn't it `nodeProfiles`?

e2e/aws_test.go
```diff
@@ -2,13 +2,70 @@ package e2e
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 )
 
 var _ = framework.RoboDescribe("AWS Integration Test", func() {
 	f := framework.New()
+	ctx := framework.TestContext
+
+	It("should provision a new cluster [provisioner:aws_install]", func() {
```
alex-kovoy: yes, to give a structure. We tried this format  [provisioner:aws] [install][expand]
but decided to replace  [] with _ so you do not need to escape it with \\ when running focused tests. 
-  \\[provisioner:aws\\] \\[install\\]\\[expand\\]

PR #71, opened 2017-06-12
misc updates to sanity tests

DenisMishin:


suite/suite_test.go
```diff
@@ -26,6 +48,22 @@ var suites = map[string]testSet{
 	"sanity": sanity.Basic,
 }
 
+var storageDriverOsCompat = map[string][]string{
+	"ubuntu": []string{"overlay2", "overlay", "devicemapper", "loopback"},
+	"debian": []string{"overlay2", "overlay", "devicemapper", "loopback"},
+	"centos": []string{"overlay2", "overlay", "devicemapper", "loopback"},
+	"rhel":   []string{"devicemapper", "loopback"},
```
sofuture: Why not `overlay` + `overlay2`. IIRC both are supported.

DenisMishin: as per https://github.com/gravitational/gravity/issues/2254

also:
```
OverlayFS remains a Technology Preview in Red Hat Enterprise Linux 7.2 under most circumstances. As such, the kernel will log warnings when this technology is activated.
Full support is available for OverlayFS when used with Docker under the following restrictions:
* OverlayFS is only supported for use as a Docker graph driver. Its use can only be supported for container COW content, not for persistent storage. Any persistent storage must be placed on non-OverlayFS volumes to be supported. Only default Docker configuration can be used; that is, one level of overlay, one lowerdir, and both lower and upper levels are on the same file system.
* Only XFS is currently supported for use as a lower layer file system.
* SELinux must be enabled and in enforcing mode on the physical machine, but must be disabled in the container when performing container separation; that is, /etc/sysconfig/docker must not contain --selinux-enabled. SELinux support for OverlayFS is being worked on upstream, and is expected in a future release.
* The OverlayFS kernel ABI and userspace behavior are not considered stable, and may see changes in future updates.
* In order to make the yum and rpm utilities work properly inside the container, the user should be using the yum-plugin-ovl packages.
```

PR #75, opened 2017-06-27
dynamic configs and ops registration

DenisMishin:
* moved all hardcoded configs out of tests, CLI configurable now
* option to register an installation with remote ops center


lib/config/config.go
```diff
@@ -0,0 +1,141 @@
...
+func (c *Config) Add(key string, fn ConfigFn, defaults interface{}) {
+	c.entries[key] = entry{fn, defaults}
+}
+
+// ParseAndConfigure will take list of function=JSON, base config map, and return list of initialized test functions to run
+func (c *Config) Parse(args []string) (fns map[string]Entry, err error) {
+	var errs []error
+	fns = map[string]Entry{}
+
+	for _, arg := range args {
```
DenisMishin: VisitAll is for flags, i'm walking flag.Args()

PR #94, opened 2017-11-15
Bootstrap fixes

UnderGreen:


infra/terraform/file_transfer.go
```diff
@@ -56,12 +56,13 @@ func (t *terraform) makeRemoteCommand(fileUrl, command string) (string, error) {
 			do test -f /var/lib/bootstrap_complete && break || \
 			echo Waiting for bootstrap to complete && sleep 15 ; \
 		done &&  \
-		echo Cleaning up && rm -rf %[1]s/installer/* && 
-		echo Downloading installer %[5]s to %[3]s ... && %[2]s && \
+		echo Cleaning up && rm -rf %[1]s/installer/* && \
+        	if [[ ! -f %[3]s ]]; then echo Downloading installer %[5]s to %[3]s ... ; %[2]s; fi && \
```
UnderGreen: Do not download installer, because it can be exists after executing client's bootstrap. Inside `user_data` for example.

infra/terraform/terraform.go
```diff
@@ -270,6 +270,7 @@ func (r *terraform) boot(ctx context.Context) (output string, err error) {
 		"-var", fmt.Sprintf("nodes=%d", r.NumNodes),
 		"-var", fmt.Sprintf("os=%s", r.OS),
 		"-var", fmt.Sprintf("random_password=%s", uuid.NewV4().String()),
+		"-var", fmt.Sprintf("installer_url=%s", r.InstallerURL),
```
UnderGreen: Pass installer_url variable to terraform, because it can be used in client's bootstrap script.

DenisMishin: Why should one transfer installer using bootstrap script? 
robotest has sufficient support for all key installer transfer methods (s3, http and localfile)

PR #100, opened 2017-11-27
Kevin/ops center

knisbet:
Addresses https://github.com/gravitational/gravity/issues/2977

Adds support to robotest to test autoscaling clusters that support autoscaling groups in AWS and testing https://github.com/gravitational/provisioner when a PR is created.
- Create a new cloud type "ops", which will use an ops center to provision a telekube cluster instead of directly interacting with the cloud provider
- Create a new test autoscale, which will interact with the AWS API to scale up/down the cluster

See https://jenkins.gravitational.io/job/ProvisionerCI/configure for the Jenkins CI flow

I don't have much profession go experience, so feedback welcome.


klizhentas:
@a-palchikov @DenisMishin please review chaps (added Dima as well).

infra/gravity/provision.go
```diff
@@ -57,8 +66,151 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+			}
+
+			switch status {
+			case "installing":
+				// we're still installing, just continue the loop
+			case "active":
+				// the cluster install completed, we can continue the install process
+				break Loop
+			default:
+				return nil, destroyFn, trace.Wrap(fmt.Errorf("unexpected cluster status: %v", status))
```
a-palchikov: `trace.BadParameter` instead (and everywhere else using `fmt.Errorf`)

infra/gravity/provision.go
```diff
@@ -57,8 +66,151 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+	ctx, cancel := context.WithTimeout(c.Context(), cloudInitTimeout)
+	defer cancel()
+
+	c.Logger().Debugf("running post provisioning tasks")
+	err = c.postProvision(cfg, gravityNodes)
+	if err != nil {
+		return nil, destroyFn, trace.Wrap(err)
+	}
+
+	c.Logger().Debug("Ensuring disk speed is adequate across nodes")
```
a-palchikov: `ensuring`

infra/gravity/provision.go
```diff
@@ -57,8 +66,151 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+	ctx, cancel = context.WithTimeout(c.Context(), diskWaitTimeout)
+	defer cancel()
+	err = waitDisks(ctx, gravityNodes, []string{"/iotest", path.Join(cfg.dockerDevice, "/iotest")})
+	if err != nil {
+		err = trace.Wrap(err, "VM disks do not meet minimum write performance requirements")
+		c.Logger().WithError(err).Error(err.Error())
+		return nil, destroyFn, err
```
a-palchikov: `trace.Wrap(err)`

infra/gravity/provision.go
```diff
@@ -57,8 +66,151 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+	}
+
+	c.Logger().Info("provisioning complete")
+	return gravityNodes, destroyFn, nil
+}
+
+// Provision gets VMs up, running and ready to use
```
a-palchikov: s/Provision/provisionCloud

infra/gravity/provision.go
```diff
@@ -113,6 +253,27 @@ func (c *TestContext) Provision(cfg ProvisionerConfig) ([]Gravity, DestroyFn, er
 	return gravityNodes, wrapDestroyFn(c, cfg.Tag(), gravityNodes, destroyFn), nil
 }
 
+// postProvision runs common tasks for both ops and cloud provisioners once the VM's have been setup and are running
```
a-palchikov: `VMs`

infra/gravity/provision.go
```diff
@@ -113,6 +253,27 @@ func (c *TestContext) Provision(cfg ProvisionerConfig) ([]Gravity, DestroyFn, er
...
+func (c *TestContext) postProvision(cfg ProvisionerConfig, gravityNodes []Gravity) error {
+	c.Logger().Debug("Streaming logs")
```
a-palchikov: lower-case logs - elsewhere too

infra/gravity/provision.go
```diff
@@ -207,6 +368,9 @@ func configureVM(ctx context.Context, log logrus.FieldLogger, node infra.Node, p
 		err = bootstrapAWS(ctx, g, param)
 	case "azure":
 		err = bootstrapAzure(ctx, g, param)
+	case "ops":
+		// For ops installs, we don't run the installer. So just hardcode the install directory to /bin
+		g.installDir = "/bin"
```
a-palchikov: can it be left empty?

infra/gravity/provision_test.go
```diff
@@ -0,0 +1,45 @@
+package gravity
+
+import (
+	"testing"
+
+	"github.com/gravitational/robotest/infra"
+	log "github.com/sirupsen/logrus"
+)
+
+func TestGenerateOpsClusterDefn(t *testing.T) {
```
a-palchikov: `Definition`?

infra/gravity/provision_test.go
```diff
@@ -0,0 +1,45 @@
...
+  aws:
+    keyName: ops
+    region: us-east-2
+  nodes:
+  - profile: node
+    count: 5
+    instanceType: c4.2xlarge
+  provider: aws` {
+		t.Error("generated cluster configuration doesn't match expected output")
+		log.Info(defn)
```
a-palchikov: I think `t.Log` is more appropriate

infra/gravity/terraform.go
```diff
@@ -169,19 +169,22 @@ func makeDynamicParams(baseConfig ProvisionerConfig) (*cloudDynamicParams, error
...
 		},
+		"ops": map[string]string{
+			"centos": "centos",
+		},
 	}
 
 	param.user, ok = usernames[baseConfig.CloudProvider][baseConfig.os.Vendor]
 	if !ok {
-		return nil, trace.BadParameter(baseConfig.os.Vendor)
+		return nil, trace.BadParameter("os.Vendor not found: %v", baseConfig.os.Vendor)
```
a-palchikov: `unknown OS vendor: %q`

lib/constants/constants.go
```diff
@@ -48,4 +48,7 @@ const (
 
 	// Azure is microsoft azure cloud
 	Azure = "azure"
+
+	// Ops is ops center
```
a-palchikov:
```go
Ops specifies a special cloud provider - a telekube Ops Center
```

suite/sanity/autoscale.go
```diff
@@ -0,0 +1,31 @@
+package sanity
+
+import (
+	"github.com/gravitational/robotest/infra/gravity"
+)
+
+// resize installs an initial cluster and then expands or gracefully shrinks it to given number of nodes
```
a-palchikov: s/resize/autoscale

suite/suite_test.go
```diff
@@ -101,6 +101,8 @@ func setupSignals(suite gravity.TestSuite) {
 // right now it'll just invoke sanity suite
 func TestMain(t *testing.T) {
 	if *testSuite == "" || *tag == "" {
+		logrus.Error("testSuite: ", testSuite)
```
a-palchikov: Are these debugging left-overs?

infra/gravity/provision.go
knisbet: Not easily, the gravity command appears to be called as follows:
```
cmd := fmt.Sprintf("cd %s && sudo ./gravity status --system-log-file=./telekube-system.log", g.installDir)
```

So unless I try and refactor that code to handle the case where installDir isn't set to try and run gravity from the path, installDir needs to get set somewhere.

infra/gravity/ops_center.go
```diff
@@ -0,0 +1,160 @@
...
+	return func() error {
+		log := tc.Logger().WithFields(logrus.Fields{
+			"cluster": clusterName,
+		})
+
+		// if the TestContext has an error, it means this robotest run failed
+		if tc.Context().Err() != nil {
+			log.WithError(tc.Context().Err()).Info("error caught in testing")
+
+			if policy.DestroyOnFailure == false {
```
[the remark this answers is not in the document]
knisbet: fixed. Set both to use ==

infra/gravity/ops_center.go
```diff
@@ -0,0 +1,160 @@
...
+		if err != nil {
+			return err
+		}
+		log.Info("tele rm result: ", string(out))
+
+		// monitor the cluster until it's gone
+		timeout := time.After(DefaultTimeouts.Uninstall)
+		tick := time.Tick(5 * time.Second)
+
+		for {
```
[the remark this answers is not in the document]
knisbet: It does, the only thing is retryer is based on delay + attempts, where as I'm reusing DefaultTimeouts.Uninstall which is based on an absolute timeout.

knisbet:
@a-palchikov I think I addressed everything you noticed. Thank You!

infra/gravity/ops_center.go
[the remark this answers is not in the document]
a-palchikov: We could use [backoff](https://github.com/cenkalti/backoff) to replace all retryer logic and the use of attempts with flexible backoff implementations.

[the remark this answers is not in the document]
klizhentas: I think in this case there is nothing wrong with simple ticker

infra/gravity/autoscaling.go
```diff
@@ -0,0 +1,134 @@
...
+
+// checkForNodeAssignment will check to see if our auto-scaling group has the requested number of nodes assigned
+func checkForNodeAssignment(svc *autoscaling.AutoScaling, describeASG *autoscaling.DescribeAutoScalingGroupsInput, target int) (*autoscaling.DescribeAutoScalingGroupsOutput, error) {
+	result, err := svc.DescribeAutoScalingGroups(describeASG)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	if len(result.AutoScalingGroups) != 1 {
+		return nil, trace.Errorf("unexpected number of autoscaling groups found: 1 != %v", len(result.AutoScalingGroups))
```
klizhentas: use trace.BadParameter in this and other cases as it's more explicit type

infra/gravity/ops_center.go
```diff
@@ -0,0 +1,159 @@
+package gravity
```
klizhentas: let's name the file `opscenter.go` as we don't use underscores in file names except tests

infra/gravity/ops_center.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/gravitational/trace"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	errClusterNotFound = errors.New("cluster not found")
```
klizhentas: please use explicit trace.NotFoundError instead of custom variable here and in similar cases

infra/gravity/ops_center.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		logrus.WithError(err).Error("unable to parse tele get clusters: ", string(out))
+		return "", trace.WrapWithMessage(err, string(out))
+	}
+	return res, nil
+}
+
+// parseClusterStatus will look at the command output, and search for the string "status: abc" and return
+// the "abc" portion.
+func parseClusterStatus(clusterName string, rd io.Reader) (string, error) {
```
klizhentas: why not use yaml or json parser? this logic seems error prone, I would replace it with proper parser.

infra/gravity/ops_center.go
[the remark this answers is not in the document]
klizhentas: I think in our code base you would find ! as a more frequent notation, so I propose to stick to ! for booleans

infra/gravity/provision.go
```diff
@@ -57,8 +68,152 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+		return nil, nil, trace.Wrap(err)
+	}
+	c.Logger().Debug("logging into the ops center")
+	out, err := exec.Command("tele", "login", "-o", cfg.Ops.URL, "--key", cfg.Ops.OpsKey).CombinedOutput()
+	if err != nil {
+		return nil, nil, trace.WrapWithMessage(err, string(out))
+	}
+
+	// generate a random cluster name
+	clusterName := fmt.Sprint("robotest-", uuid.NewV4().String())
```
klizhentas: would be helpful to add test case name prefix to make it easier to identify the tests

infra/gravity/provision.go
```diff
@@ -57,8 +68,152 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+
+	c.Logger().Debug("validating configuration")
+	err = validateConfig(cfg)
+	if err != nil {
+		return nil, nil, trace.Wrap(err)
+	}
+
+	c.Logger().Debug("generating ops center cluster configuration")
+	clusterPath := path.Join(cfg.StateDir, "cluster.yaml")
+	err = os.MkdirAll(cfg.StateDir, 0755)
```
klizhentas: seems inconsistent - you use constant for mask below, but not here

infra/gravity/ops_center.go
klizhentas: also, can you add license header

infra/gravity/ops_center.go
knisbet: I was a little torn on this. The options I see are:

- Just try and parse the text
- duplicate the object structure into robotest, which would then open the possibility for it to diverge
- vendor gravity, and get the structs from the gravity source code

So when I looked at other sections of robotest, it looks like it was doing parsing, such as https://github.com/gravitational/robotest/blob/f9ff9f4c8a3ddaceacbc84c2438dbe26f8953364/infra/gravity/parse.go#L18

Anyway's, I'm not opposed to vendoring in gravity so I can use the existing object definitions if you think that's the best approach.

klizhentas: I was thinking actually about unmarshalling the whole thing into substructure having only the fields you need, and it will be OK for now.

infra/gravity/ops_center.go
knisbet: Re: underscores My bad, there are other files in infra/gravity that have underscores so I was just following what I saw.

Re: license header. Do we have a reference header documented anywhere? It looks like it's missing consistently throughout the project.

infra/gravity/provision.go
```diff
@@ -57,8 +68,152 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+			default:
+				return nil, destroyFn, trace.BadParameter("unexpected cluster status: %v", status)
+			}
+		}
+	}
+
+	// now that the requested cluster has been created, we have to build the []Gravity slice of nodes
+	c.Logger().Debug("attempting to get a listing of instances from AWS for the cluster")
+	ec2svc := ec2.New(sess)
+	gravityNodes, err := c.getAWSNodes(ec2svc, "tag:KubernetesCluster", clusterName)
```
a-palchikov: `err` is ignored

infra/ops/node.go
```diff
@@ -0,0 +1,59 @@
+package ops
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/gravitational/robotest/infra"
+	"github.com/gravitational/trace"
+	"golang.org/x/crypto/ssh"
+  sshutils "github.com/gravitational/robotest/lib/ssh"
```
a-palchikov: something's up with the indent

infra/gravity/ops_center.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+	}
+
+	return "", trace.BadParameter("invalid input")
+}
+
+// generateClusterConfig will generate a cluster configuration for the ops center based
+// on the built in template
+func generateClusterConfig(cfg ProvisionerConfig, clusterName string) (string, error) {
+	template, err := template.New("cluster").Parse(`
```
a-palchikov: I would still move this to init block (make it a package global) - so that the parse errors can be caught as early as possible:
```go
var clusterConfigTemplate = template.Must(template.New("cluster").Parse(`...`))
```

infra/gravity/ops_center_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"strings"
+	"testing"
+)
+
+func TestParseClusterStatus(t *testing.T) {
+	t.Run("NotFound", func(t *testing.T) {
+		txt := `cluster kevin-ci.22.7 not found`
+
+		status, err := parseClusterStatus("kevin-ci.22.7", strings.NewReader(txt))
+		if err.Error() != errClusterNotFound.Error() {
```
a-palchikov: this could become
```go
if !trace.IsNotFound(err) {
	t.Error("expected not found error:", err)
}
```
you don't really need `errClusterNotFound`

infra/gravity/autoscaling.go
a-palchikov: maybe return `trace.Retry` as this is used inside a retry loop?

klizhentas:
@knisbet @a-palchikov ping chaps, can we get this merged?

infra/gravity/provision.go
knisbet: fixed, thanks!

infra/gravity/opscenter.go
```diff
@@ -0,0 +1,164 @@
...
+		timeout := time.After(DefaultTimeouts.Uninstall)
+		tick := time.Tick(5 * time.Second)
+
+		for {
+			select {
+			case <-timeout:
+				return trace.LimitExceeded("clusterDestroy timeout exceeded")
+			case <-tick:
+				// check provisioning status
+				status, err := getTeleClusterStatus(clusterName)
```
DenisMishin: any external commands need be executed with context.Context passed. it's no fun to debug robotest with tons of concurrent tests running and some external program got hung.

infra/gravity/provision.go
```diff
@@ -57,8 +68,155 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+	out, err = exec.Command("tele", "create", clusterPath).CombinedOutput()
+	if err != nil {
+		return nil, nil, trace.Wrap(err, string(out))
+	}
+
+	// destroyFn defines the clean up function that will destroy provisioned resources
+	destroyFn := cfg.DestroyOpsFn(c, clusterName)
+
+	// monitor the cluster until it's created or times out
+	timeout := time.After(cloudInitTimeout)
```
DenisMishin:
```
ctx, cancel := context.WithTimeout(parentContext, withDuration(c.timeouts.Install, len(nodes))
defer cancel()
```

infra/gravity/provision.go
```diff
@@ -57,8 +68,155 @@ func configureVMs(baseCtx context.Context, log logrus.FieldLogger, params cloudD
...
+	tick := time.Tick(15 * time.Second)
+
+	c.Logger().Debug("waiting for ops center provisioning to complete")
+Loop:
+	for {
+		select {
+		case <-timeout:
```
DenisMishin: use Retryer, that would make code more consistent overall

infra/gravity/autoscaling.go
```diff
@@ -0,0 +1,134 @@
...
+		return trace.Wrap(err)
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	ec2svc := ec2.New(sess)
+	for _, instance := range result.AutoScalingGroups[0].Instances {
+		// attempt to get the actual instance for each instance-id in the cluster
+		node, err := c.getAWSNodes(ec2svc, "instance-id", *instance.InstanceId)
```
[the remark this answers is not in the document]
knisbet: I'm not sure that I agree that this warrants the additional complexity of trying to make this parallel. 

In this case, the bootstrap really happens by the AWS built in provisioning when the nodes boots. All this code does is really SSH to each node, in sequence, which relatively speaking is a fast operation, and we have to wait for all nodes to complete anyway's.

infra/gravity/autoscaling.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, trace.BadParameter("unexpected number of autoscaling groups found: 1 != %v", len(result.AutoScalingGroups))
+	}
+	if len(result.AutoScalingGroups[0].Instances) != target {
+		return nil, trace.BadParameter("unexpected autoscaling count of instances. expected: %v got: %v", target, len(result.AutoScalingGroups[0].Instances))
+	}
+	return result, nil
+}
+
+// getAWSNodes will connect to the AWS API, and get a listing of nodes matching the specified filter.
+func (c *TestContext) getAWSNodes(ec2svc *ec2.EC2, filterName string, filterValue string) ([]Gravity, error) {
```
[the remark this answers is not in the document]
knisbet: I'm not sure bootstrapAWSInstance makes sense here, mainly because there is already a bootstrapAWS function that is used when the cloud provider is set to AWS, that has very different usage. So I'm not sure I really consider this bootstrap, as it's not consistent with the way bootstrapXXX is used for AWS/Azure.

infra/gravity/autoscaling.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	var nodes []Gravity
+	for _, reservation := range resp.Reservations {
+		for _, inst := range reservation.Instances {
+			node := ops.New(*inst.PublicIpAddress, *inst.PrivateIpAddress, c.provisionerCfg.Ops.SSHUser, c.provisionerCfg.Ops.SSHKeyPath)
+
+			gravityNode, err := configureVM(c.Context(), c.Logger(), node, *cloudParams)
```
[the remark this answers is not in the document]
knisbet: Sorry I'm not sure I understand the suggestion

PR #101, opened 2017-11-28
support custom script post installer transfer

DenisMishin:
fixes https://github.com/gravitational/gravity/issues/3003

### DIFF
--- a/infra/gravity/node_commands.go
+++ b/infra/gravity/node_commands.go
@@ -27,6 +27,8 @@ type Gravity interface {
 	json.Marshaler
 	// SetInstaller transfers and prepares installer package
 	SetInstaller(ctx context.Context, installerUrl string, subdir string) error
+	// ExecScript transfers and executes script with predefined parameters
+	ExecScript(ctx context.Context, scriptUrl string, args []string) error
 	// Install operates on initial master node
 	Install(ctx context.Context, param InstallParam) error
 	// Status retrieves status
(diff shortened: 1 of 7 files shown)


infra/gravity/node_commands.go
```diff
@@ -378,6 +380,25 @@ func (g *gravity) SetInstaller(ctx context.Context, installerUrl string, subdir
...
+
+	spath, err := sshutils.TransferFile(ctx, g.Client(), log,
+		scriptUrl, defaults.TmpDir, g.param.env)
+	if err != nil {
+		log.WithError(err).Error("failed to transfer script")
+		return trace.Wrap(err)
+	}
+
+	err = sshutils.Run(ctx, g.Client(), log,
+		fmt.Sprintf("sudo /bin/bash -x %s %s", spath, strings.Join(args, " ")), nil)
```
UnderGreen: For why use debug(-x) here? And why bash, not sh... maybe better to copy script, chmod +x on the node end execute. Inside script we can use preferred shell via shebang.

suite/README.md
```diff
@@ -116,6 +116,15 @@ Inherits parameters from `install`, plus:
 
 `replace_variety` will generate a combination of `replace` parameterized tests.
 
+### Post installer transfer script
```
UnderGreen: Post bootstrap?

a-palchikov: Maybe
```
### Post Installer Bootstrap Script
```

infra/gravity/node_commands.go
DenisMishin: cause otherwise if your script would fail, you'll have no clue why. I've seen it many times with bootstrap scripts which silently quit and you have no idea which program caused that cryptic error message.

`bash` because it's available on all VMs we use, and it has some richer syntax.

suite/README.md
DenisMishin: no it's specifically post installer transfer.

PR #103, opened 2017-12-05
Kevin/ops center 2

knisbet:
Oops, I thought I had already opened a PR for this, but I guess I forgot about it when waiting for the tests to complete. 

This is a part 2 to #100 to fix some of Denis's suggestions. I didn't try to parallelize any of the operations, as I don't believe there is much benefit here, as all the provisioning happens in the ops center, and the local operations only really establish SSH sessions to the nodes, which has a relatively small time cost.

### DIFF
--- a/infra/gravity/opscenter.go
+++ b/infra/gravity/opscenter.go
@@ -3,7 +3,6 @@ package gravity
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"os/exec"
 	"strings"
 	"text/template"
@@ -22,12 +21,8 @@ import (
 func getTeleClusterStatus(ctx context.Context, clusterName string) (string, error) {
 	out, err := exec.CommandContext(ctx, "tele", "get", "clusters", clusterName, "--format", "yaml").Output()
 	if err != nil {
-		// tele get clusters will get a non zero exit status code if the cluster is not found, but we want to make sure
-		// it maps to a NOT found error if the cluster isn't found
-		if out != nil {
-			if strings.Contains(string(out), fmt.Sprintf("cluster %v not found", clusterName)) {
-				return "", trace.NotFound("cluster not found")
-			}
+		if isClusterNotFoundError(out) {
+			return "", trace.NotFound("cluster not found")
 		}
 
 		logrus.WithError(err).Error("unable to parse tele get clusters: ", string(out))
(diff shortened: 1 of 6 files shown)


infra/gravity/opscenter.go
```diff
@@ -2,23 +2,34 @@ package gravity
...
 
 // getTeleClusterStatus will attempt to get the cluster status from an ops center
 // for now, it's just looking for the specific line of output from tele get clusters
 // but could parse the entire yaml structure if imported from gravity
-func getTeleClusterStatus(clusterName string) (string, error) {
-	out, err := exec.Command("tele", "get", "clusters", clusterName, "--format", "yaml").Output()
+func getTeleClusterStatus(ctx context.Context, clusterName string) (string, error) {
+	out, err := exec.CommandContext(ctx, "tele", "get", "clusters", clusterName, "--format", "yaml").Output()
 	if err != nil {
+		// tele get clusters will get a non zero exit status code if the cluster is not found, but we want to make sure
```
a-palchikov:
```go
"tele get clusters" returns a non-zero exit code ...
```
I'd rather have this documented with a function instead of the comment:
```go
if isClusterNotFoundError(out) {
	return "", trace.NotFound("cluster not found")
}
```

infra/gravity/opscenter.go
```diff
@@ -127,38 +141,39 @@ func (c ProvisionerConfig) DestroyOpsFn(tc *TestContext, clusterName string) fun
...
+				// de-provisioning completed
+				return nil
 			}
+			if err != nil {
+				return trace.Wrap(err)
+			}
+
+			switch status {
+			case "uninstalling":
+				return trace.BadParameter("still uninstalling")
```
a-palchikov:
```go
return trace.Retry("still uninstalling")
```

infra/gravity/provision.go
```diff
@@ -155,46 +154,47 @@ func (c *TestContext) provisionOps(cfg ProvisionerConfig) ([]Gravity, DestroyFn,
...
+		status, err := getTeleClusterStatus(ctx, clusterName)
+		if err != nil {
+			return trace.Wrap(err)
 		}
+		c.Logger().WithField("status", status).Debug("provisioning status")
+
+		switch status {
+		case "installing":
+			// we're still installing, just continue the loop
+			return trace.BadParameter("installation not complete")
```
a-palchikov: `trace.Retry`?
Most of the time comments are not necessary if other things agree - e.g. logging/error messages giving the same hints. Comments should add value where code is otherwise hard to understand. In other cases they're a distraction.

infra/gravity/opscenter.go
```diff
@@ -165,7 +172,7 @@ func (c ProvisionerConfig) DestroyOpsFn(tc *TestContext, clusterName string) fun
 
 			switch status {
 			case "uninstalling":
-				return trace.BadParameter("still uninstalling")
+				return trace.Retry(trace.BadParameter("uninstall not complete"), "uninstall not complete")
```
a-palchikov: You don't need to pass another error - error message should suffice:
```go
return trace.Retry(nil, "uninstall not complete")
```

PR #104, opened 2017-12-05
Execute post installer script

UnderGreen:
Because for air-gapped applications we should configure Iptables to drop all output traffic. It is impossible to do before downloading the installer, because download procedure will fail.

### DIFF
--- a/e2e/framework/test_context.go
+++ b/e2e/framework/test_context.go
@@ -258,6 +258,10 @@ type OnpremConfig struct {
 	// ScriptPath defines the path to the provisioner script.
 	// TODO: if unspecified, scripts in assets/<provisioner> are used
 	ScriptPath string `json:"script_path" yaml:"script_path"`
+	// PostInstallerScript defines the path on remote node
+	// to the script which should be called on installer node
+	// after downloading of the installer
+	PostInstallerScript string `json:"post_installer_script" yaml:"post_installer_script"`
 	// ExpandProfile specifies an optional name of the server profile for On-Premise expand operation.
 	// If the profile is unspecified, the test will use the first available.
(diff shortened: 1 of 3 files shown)


e2e/framework/test_context.go
```diff
@@ -258,6 +258,10 @@ type OnpremConfig struct {
 	// ScriptPath defines the path to the provisioner script.
 	// TODO: if unspecified, scripts in assets/<provisioner> are used
 	ScriptPath string `json:"script_path" yaml:"script_path"`
+	// PostInstallerScript defines the path on remote node
```
a-palchikov:
```
PostInstallerScript defines a path to the script on a remote node that is executed after the installer has been downloaded
```

infra/terraform/file_transfer.go
```diff
@@ -60,8 +60,10 @@ func (t *terraform) makeRemoteCommand(fileUrl, command string) (string, error) {
 		echo Downloading installer %[5]s to %[3]s ... && %[2]s && \
 		echo Creating installer dir && mkdir -p %[1]s/installer && \
 		echo Unpacking installer && tar -xvf %[3]s -C %[1]s/installer && \
+		echo Checking existence of post-downloading installer script and executing it && \
```
a-palchikov: Can you change this to a `text/tempate` - I'm not comfortable with more than 3 unnamed arguments.

UnderGreen:
Because for air-gapped applications we should configure Iptables to drop all output traffic. It is impossible to do before downloading the installer, because download procedure will fail.

infra/terraform/file_transfer.go
UnderGreen: Ok, will do

PR #107, opened 2017-12-20
Updated e2e Dockerfile and Makefile

UnderGreen:


UnderGreen:
@a-palchikov both workflows use the same `TERRAFORM_VERSION` from `docker/Makefile`

PR #115, opened 2018-02-27
Custom terraform vars script

UnderGreen:


infra/terraform/terraform.go
```diff
@@ -314,7 +323,7 @@ func (r *terraform) command(ctx context.Context, args []string, opts ...system.C
 		))
 	err := system.ExecL(cmd, &out, r.Entry, opts...)
 	if err != nil {
-		return out.Bytes(), trace.Wrap(err, "command %q failed (args %q, wd %q)", cmd.Path, cmd.Args, cmd.Dir)
+		return out.Bytes(), trace.Wrap(err, "command %q failed (args %q, wd %q), stdin/stdout: %v", cmd.Path, cmd.Args, cmd.Dir, out.String())
```
a-palchikov: I think this should be `stdout/stderr` but even this is superfluous. Why not simply output it after a semicolon?

e2e/framework/test_context.go
```diff
@@ -282,6 +282,8 @@ type OnpremConfig struct {
 	DockerDevice string `json:"docker_device" yaml:"docker_device" validate:"required"`
 	// ClusterAddress defines configuration for accessing installed cluster web page
 	ClusterAddress *ClusterAddress `json:"cluster_address" yaml:"cluster_address"`
+	// CustomVarsFile defines the path to file with custom terraform variables
+	CustomVarsFile string `json:"custom_vars_file" yaml:"custom_vars_file"`
```
a-palchikov: `VariablesFile` or `VariablesPath` would be a better fit.

infra/terraform/terraform.go
```diff
@@ -199,12 +199,16 @@ func (r *terraform) Destroy(ctx context.Context) error {
...
-	_, err := r.command(ctx, []string{
+	destroyCommand := []string{
 		"destroy", "-force",
 		"-var", fmt.Sprintf("nodes=%d", r.NumNodes),
 		"-var", fmt.Sprintf("os=%s", r.OS),
 		fmt.Sprintf("-var-file=%s", varsPath),
-	})
+	}
+	if r.CustomVarsFile != "" {
+		destroyCommand = append(destroyCommand, fmt.Sprintf("-var-file=%s", r.CustomVarsFile))
```
knisbet: Is it intentional that -var-file get's passed twice? Once with varsPath and once with r.CustomVarsFile?

infra/terraform/terraform.go
```diff
@@ -289,13 +293,18 @@ func (r *terraform) boot(ctx context.Context) (output string, err error) {
...
+	applyCommand := []string{
 		"apply", "-input=false", "-auto-approve",
 		"-var", fmt.Sprintf("nodes=%d", r.NumNodes),
 		"-var", fmt.Sprintf("os=%s", r.OS),
 		"-var", fmt.Sprintf("random_password=%s", uuid.NewV4().String()),
 		fmt.Sprintf("-var-file=%s", varsPath),
-	})
+	}
+	if r.CustomVarsFile != "" {
+		applyCommand = append(applyCommand, fmt.Sprintf("-var-file=%s", r.CustomVarsFile))
```
knisbet: Same?

infra/terraform/terraform.go
UnderGreen: Yes. `varsPath` contains variables which we get from robotest config, like aws user/pass and etc. `r.CustomVarsFile` provides a way to add custom variables to terraform command, without adding all that variables into robotest config.

PR #118, opened 2018-04-04
Fix the terraform data attribute for dynamic IP.

a-palchikov:
* Fix the terraform data attribute for dynamic IP.
 * Fix another place where resource destruction could be subject to a soon expiring context - use a fresh context instead.


a-palchikov:
Let me test this properly first.

PR #124, opened 2018-05-23
Updated Terraform and related providers to latest release

eldios:
Updated Terraform and related providers to latest release


eldios:
@klizhentas I'm gonna recreate this and the branch since this was not the commit with the information leak, this was the already fixed one.

Anyway, thanks for the prompt action :+1:

PR #134, opened 2018-11-13
Dmitri/3346 gce

a-palchikov:
This PR adds support for `GCE` as infrastructure target for robotest.


assets/terraform/azure/config.tf
```diff
@@ -2,26 +2,30 @@
 # Uses Azure Resource Manager provider
 #   https://www.terraform.io/docs/providers/azurerm
 # 
-variable "subscription_id"{ }
-variable "client_id"      { }
-variable "client_secret"  { }
-variable "tenant_id"      { }
+variable "subscription_id" {}
```
a-palchikov: Nothing's changed here - just `terraform fmt`.

assets/terraform/gce/output.tf
```diff
@@ -0,0 +1,20 @@
...
+
+output "private_ips" {
+  value = ["${google_compute_instance.node.*.network_interface.0.address}"]
+}
+
+output "public_ips" {
+  value = ["${google_compute_instance.node.*.network_interface.0.access_config.0.assigned_nat_ip}"]
+}
+
+# output "pod_cidr_blocks" {
```
a-palchikov: I want to keep these (and all related commented out block elsewhere) as these can be used to our advantage when there's a fannel alternative as we can save some route quotas.

e2e/framework/applications.go
```diff
@@ -84,10 +85,10 @@ func BackupApplication() {
 	Expect(TestContext.Extensions.BackupConfig.Path).NotTo(BeNil(), "expect valid path to backup file")
 
 	backupNode, err := Cluster.Provisioner().NodePool().Node(TestContext.Extensions.BackupConfig.Addr)
-	if err != nil {
-		trace.NotFound("node with address %v not found in config state", TestContext.Extensions.BackupConfig.Addr)
-	}
-	Distribute(fmt.Sprintf("sudo gravity planet enter -- --notty /usr/bin/gravity -- system backup %s %s", TestContext.Application.String(), TestContext.Extensions.BackupConfig.Path), backupNode)
+	Expect(err).NotTo(HaveOccurred(),
```
a-palchikov: I had to fix a bunch of code as I introduced a linting step to the makefile.

e2e/framework/framework.go
```diff
@@ -140,11 +141,17 @@ func InitializeCluster() {
 			Expect(err).NotTo(HaveOccurred())
 
 			withInstaller := TestContext.Onprem.InstallerURL != "" && TestContext.Wizard
-			if TestContext.Provisioner.LoadFromState {
-				installerNode, err = provisioner.LoadFromState(TestContext.Provisioner.StateFile, withInstaller)
+			loader, ok := provisioner.(infra.ExternalStateLoader)
```
a-palchikov: Not directly related to this PR but it's something I'd put in in the beginning to avoid having a `trace.NotImplemented` for `vagrant` as the ability to load state from external state file is only relevant for `terraform` provisioner (i.e. `terraform.tfstate`)

e2e/uimodel/user/login.go
```diff
@@ -15,10 +12,9 @@ import (
 )
 
 const (
-	WithEmail         = "email"
-	WithGoogle        = "google"
-	WithNoProvider    = ""
-	googlePageTimeout = 1 * time.Second
```
a-palchikov: [lint] that includes dead code too.

infra/gravity/cluster_status.go
```diff
@@ -17,7 +17,7 @@ func (c *TestContext) Status(nodes []Gravity) error {
 	defer cancel()
 
 	retry := wait.Retryer{
-		Attempts: 1000,
+		Attempts: 100,
```
a-palchikov: I thought this was too long a delay to wait for status.

infra/gravity/node_commands.go
```diff
@@ -117,15 +122,42 @@ type JoinCmd struct {
...
 }
 
-// GravityStatus is serialized form of `gravity status` CLI.
+// GravityStatus describes the status of the Gravity cluster
 type GravityStatus struct {
-	Application string
-	Cluster     string
-	Status      string
+	// Cluster describes the cluster status
+	Cluster ClusterStatus `json:"cluster"`
```
a-palchikov: I also changed status query to use the `json` output instead of `regexp`s.

infra/gravity/provision.go
```diff
@@ -189,99 +190,98 @@ Loop:
...
 
 	err = c.postProvision(cfg, gravityNodes)
 	if err != nil {
-		c.Logger().WithError(err).Error("Post-provisioning failed, tear down as non-usable.")
-		return nil, nil, trace.Wrap(err)
+		log.WithError(err).Error("Post-provisioning failed, tear down as non-usable.")
+		return cluster, nil, trace.Wrap(err)
 	}
 
-	c.Logger().Debug("ensuring disk speed is adequate across nodes")
```
a-palchikov: This seems to have only affected `Azure` as far as I remember.

infra/gravity/testcontext.go
```diff
@@ -72,23 +81,59 @@ func (c *TestContext) Error() error {
 	return c.err
 }
 
-// Checkpoint marks milestone within a test
+// WithFields assigns additional logging fields to this context
+func (c *TestContext) WithFields(fields logrus.Fields) *TestContext {
```
a-palchikov: I wanted to this to be able to add custom logging attributes (i.e. specify the node tag the test is using as it is different from the cluster name to due how GCE resources are structured.
There's no structural hierarchy as on Azure where resources could be grouped under `Resource Group`s - instead, everything is flat. Additionally, resource names are subject to RFC1035 which limits the length to 63 characters.

lib/ssh/scp.go
```diff
@@ -51,7 +51,8 @@ func PutFile(ctx context.Context, client *ssh.Client, log logrus.FieldLogger, sr
 	}
 
 	go func() {
-		err := scpSendFile(stdin, f, fi)
+		buf := make([]byte, 32<<20)
```
a-palchikov: By default, `io.Copy` uses 32KiB intermediate buffer. This is a yet unjustified attempt to speed up (installer) file transfer.

infra/gravity/node_commands.go
```diff
@@ -308,7 +349,16 @@ func (g *gravity) Remove(ctx context.Context, node string, graceful Graceful) er
 
 // Uninstall removes gravity installation. It requires Leave beforehand
 func (g *gravity) Uninstall(ctx context.Context) error {
-	cmd := fmt.Sprintf(`cd %s && sudo ./gravity system uninstall --confirm --system-log-file=./telekube-system.log`, g.installDir)
+	cmd := fmt.Sprintf(`cd %s && sudo ./gravity leave --force --confirm --system-log-file=./telekube-system.log`, g.installDir)
```
r0mant: For multi-node cluster this will launch shrink operation rather than just cleaning local data - is that intended?

lib/ssh/scp.go
knisbet: 32<<20 is a bit of a hard read for me...

why not `1<<25`
or
```
const MiB = 1048576 
buf := make([]byte, 32*MiB)
```
?

infra/gravity/provision.go
r0mant: Azure is still supported provider for robotest, right? Maybe keep this for it at least, if it's still relevant that is.

infra/providers/gce/config.go
```diff
@@ -0,0 +1,36 @@
...
+	Zone string `json:"zone,omitempty" yaml:"zone"`
+	// VMType specifies the type of machine to provision
+	// https://cloud.google.com/compute/docs/machine-types
+	VMType string `json:"vm_type" yaml:"vm_type" validate:"required"`
+	// SSHKeyPath specifies the location of the SSH private key for remote access
+	SSHKeyPath string `json:"-" yaml:"ssh_key_path" validate:"required"`
+	// SSHUser defines SSH user to connect to the provisioned machines.
+	// Required attribute.
+	// Will be determined based on selected cloud provder.
+	SSHUser string `json:"os_user" yaml:"ssh_user"`
```
r0mant: Why are the tags different for json / yaml?

infra/gravity/node_commands.go
a-palchikov: You're right, I'll revert.

PR #136, opened 2018-11-27
Dmitri/3825 report

a-palchikov:
This PR adds an additional configuration attribute to specify the location of the tested gravity binary to upload to nodes during upgrade.
Now, robotest would only upload the update tarball to a single node it used to launch the upgrade from. After the tests, it used the gravity binaries from the base version on all other nodes to collect debug reports. This obviously has issues if report collection was improved or fixed bugs in the tested version versus the base version (the one used as a basis for upgrade).

In this PR, robotest will also upload the tested gravity binary to nodes other than the one hosting the update tarball before the test to collect debug reports using the up-to-date binary. The reason this is not instead using, for example, the binary as extracted by the agents is being able to capture logs even in the unlikely event of upgrade failing to start.
Additionally, debug logs collection was optimized to only collect kubernetes logs from a single master node to avoid redundant captures (and significantly cut down on time for collecting logs depending on the cluster size).

Fixes https://github.com/gravitational/gravity.e/issues/3825.

### DIFF
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -24,16 +24,16 @@ func (c *TestContext) Status(nodes []Gravity) error {
 	}
 
 	err := retry.Do(ctx, func() error {
-		errors := make(chan error, len(nodes))
+		errs := make(chan error, len(nodes))
 
 		for _, node := range nodes {
 			go func(n Gravity) {
 				_, err := n.Status(ctx)
-				errors <- err
+				errs <- err
 			}(node)
 		}
 
-		err := utils.CollectErrors(ctx, errors)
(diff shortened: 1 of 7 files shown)


infra/gravity/cluster_status.go
```diff
@@ -23,16 +24,16 @@ func (c *TestContext) Status(nodes []Gravity) error {
 	}
 
 	err := retry.Do(ctx, func() error {
-		errs := make(chan error, len(nodes))
+		errors := make(chan error, len(nodes))
```
knisbet: Just a personal curiosity, err is the conventional variable name for an err, so is errs not the conventional name for a list or channel of errors?

PR #138, opened 2018-12-04
Dmitri/dynamic retries

a-palchikov:
I suspect robotest getting blacklisted sometimes when running tests on ubuntu as it's running sshguard and it's possible that robotest might spam the node during bootstrapping with SSH requests and might get recognized by sshguard as hostile.
To fix this:
  * connect using a backoff interval (in general)
  * add robotest node's IP to sshguard's whitelist on Ubuntu

Additionally, current robotest behavior with the number of failed attempts fixed at 3 poses an issue with preemptible nodes. I added a new maximum to deal with preemptions which is much higher (`MaxPreemptedRetriesPerTest`) than the generic failure maximum (`MaxRetriesPerTest`). It should make the test retries more optimal in case of preemptions.


infra/gravity/provision.go
```diff
@@ -235,14 +228,23 @@ func (c *TestContext) provisionCloud(cfg ProvisionerConfig) (cluster Cluster, co
 	ctx, cancel := context.WithTimeout(c.Context(), cloudInitTimeout)
 	defer cancel()
 
-	log.Debug("Configuring VMs")
-	gravityNodes, err := configureVMs(ctx, c.Logger(), infra.params, infra.nodes)
+	log.Debug("Connecting to VMs.")
+	gravityNodes, err := connectVMs(ctx, c.Logger(), infra.params, infra.nodes)
+	if err != nil {
+		log.WithError(err).Error("Some nodes failed to connect, tear down as non-usable.")
```
r0mant: "unusable"?

infra/gravity/testcontext.go
```diff
@@ -57,6 +57,9 @@ type TestContext struct {
 	// (subject to a maximum number of retry attempts)
 	monitorCtx    context.Context
 	monitorCancel context.CancelFunc
+	// preempted indicates that a node part of the cluster that belongs
```
r0mant: "that a node part of the cluster that belongs..." <- reads weird

PR #141, opened 2019-05-29
Use full URLs for logs

a-palchikov:
Remove shortening of log URLs as the service we were using was discontinued - return full URLs at the moment. This needs to be decided later whether to migrate to Firebase Dynamic Links (i.e. stay with google) or use other URL service or do completely away with shorterning.

### DIFF
--- a/infra/gravity/testsuite.go
+++ b/infra/gravity/testsuite.go
@@ -166,10 +166,8 @@ severity>=INFO`, testUID, s.uid)},
 		}.Encode(),
 	}
 
-	// Google UIRL shortener has been discontinued.
+	// Google URL shortener has been discontinued.
 	// See https://developers.googleblog.com/2018/03/transitioning-google-url-shortener.html for details.
-	// short, err := s.client.Shorten(s.ctx, longUrl.String())
-	// return short, trace.Wrap(err)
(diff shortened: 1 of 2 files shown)


infra/gravity/testsuite.go
```diff
@@ -163,10 +163,16 @@ func (s *testSuite) getLogLink(testUID string) (string, error) {
 labels.__uuid__="%s"
 labels.__suite__="%s"
 severity>=INFO`, testUID, s.uid)},
-		}.Encode()}
+		}.Encode(),
+	}
 
-	short, err := s.client.Shorten(s.ctx, longUrl.String())
-	return short, trace.Wrap(err)
+	// Google UIRL shortener has been discontinued.
```
r0mant: URL?

infra/gravity/testsuite.go
```diff
@@ -163,10 +163,16 @@ func (s *testSuite) getLogLink(testUID string) (string, error) {
...
+	// See https://developers.googleblog.com/2018/03/transitioning-google-url-shortener.html for details.
+	// short, err := s.client.Shorten(s.ctx, longUrl.String())
```
r0mant: The service is in Google graveyard now so is there any reason to keep this code commented? I'd just keep the TODO item below.

PR #148, opened 2019-09-19
Bernard/master failover impl

bernardjkim:
- Added initial implementation for failover test.
- Need to figure out the behavior of gravity, but still having trouble actually running robotest to test it out... For the failover test, we first need to install a multi-node cluster, but I'm unable to even get robotest to pass the three node install test.


suite/sanity/failover.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/gravitational/robotest/infra/gravity"
+)
+
+// TODO
+type failoverParam struct {
+	installParam
+}
+
+// failover tests master failover
+func failover(p interface{}) (gravity.TestFunc, error) {
```
klizhentas: why is this an interface{} instead of `failoverParam`?

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	if err := oldLeader.Connect(ctx); err != nil {
+		return trace.Wrap(err, "failed to connect to cluster")
+	}
+
+	// TODO: do we need to check the status of all the nodes?
+	status, err := newLeader.Status(ctx)
+	if err != nil {
+		trace.Wrap(err)
```
klizhentas: missing `return`

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,67 @@
...
+func (c *TestContext) Failover(nodes []Gravity) error {
+	// TODO: Configure timeouts
+	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
+	defer cancel()
+
+	oldLeader, err := getLeaderNode(ctx, nodes)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	c.Logger().Info(fmt.Sprintf("oldLeader=%v", oldLeader))
```
klizhentas: better use structured logging here, e.g. `c.Logger().WithFields(log.Fields{"old": oldLeader}})`...

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return nil
+}
+
+// getLeaderNode returns the current leader node.
+func getLeaderNode(ctx context.Context, nodes []Gravity) (leader Gravity, err error) {
+	for _, node := range nodes {
+		if node.IsLeader(ctx) {
+			if leader != nil {
+				return nil, trace.BadParameter("multiple leader nodes")
```
klizhentas: should add what multiple leader nodes are found to simplify troubleshooting

infra/gravity/node_commands.go
```diff
@@ -587,6 +593,57 @@ func (g *gravity) RunInPlanet(ctx context.Context, cmd string, args ...string) (
 	return out, nil
 }
 
+// IsLeader returns true if node is leader, false o.w.
+func (g *gravity) IsLeader(ctx context.Context) bool {
+	// TODO: use planet cmd once planet cli is updated s.t. `leader view` does
+	// not require args.
+	// leaderIP, err := g.RunInPlanet(ctx, "planet", "leader", "view")
+	leaderIP, err := g.RunInPlanet(ctx, "etcdctl", "get", "gravity/local/leader")
```
klizhentas: should be a constant

infra/gravity/node_commands.go
```diff
@@ -587,6 +593,57 @@ func (g *gravity) RunInPlanet(ctx context.Context, cmd string, args ...string) (
...
+	if err == nil && leaderIP == g.Node().PrivateAddr() {
+		return true
+	}
+	return false
+}
+
+// Connect connects the node to the cluster.
```
klizhentas: perhaps `Reconnect` because it implies it was disconnected before

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,67 @@
...
+		return trace.Wrap(err, "failed to connect to cluster")
+	}
+
+	// TODO: do we need to check the status of all the nodes?
+	status, err := newLeader.Status(ctx)
+	if err != nil {
+		trace.Wrap(err)
+	}
+	// TODO: add Status.IsActive function
+	if status.Cluster.Status != "active" {
```
klizhentas: should be a constant instead of magic string

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	newLeader, err := getLeaderNode(ctx, nodes)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	c.Logger().Info(fmt.Sprintf("newLeader=%v", newLeader))
+
+	if newLeader == oldLeader {
+		return trace.BadParameter("did not failover to new leader")
```
klizhentas: add more context here, also we have new error.AddFields to add metadata

suite/sanity/failover.go
bernardjkim: Implements ConfigFn
`// ConfigFn is a function which takes default parameters and returns a pre-configured test function
 type ConfigFn func(param interface{}) (gravity.TestFunc, error)`

infra/gravity/node_commands.go
bernardjkim: Do you mean to do something like:

`etcdGetLeader := []string{"etcdctl", "get","gravity/local/leader"}`
`leaderIP, err := g.RunInPlanet(ctx, etcdGetLeader[0], etcdGetLeader[1:]...)`

or did you mean to just make the etcd key a constant?

klizhentas: more like `const LeaderKey="gravity/local/leader"`

infra/gravity/node_commands.go
```diff
@@ -587,6 +593,57 @@ func (g *gravity) RunInPlanet(ctx context.Context, cmd string, args ...string) (
...
+func (g *gravity) Connect(ctx context.Context) error {
+	status, err := g.Status(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	for _, node := range status.Cluster.Nodes {
+		// TODO: look into using existing libs for manipulating network traffic.
```
knisbet: coreos/go-iptables might be worth checking out. Here's an example from wormhole: https://github.com/gravitational/wormhole/blob/master/pkg/iptables/iptables.go

infra/gravity/node_commands.go
```diff
@@ -74,6 +74,12 @@ type Gravity interface {
 	Client() *ssh.Client
 	// Will log using extended info such as current tag, node info, etc
 	Logger() logrus.FieldLogger
+	// Connect connects the node to the cluster.
```
knisbet: Nit: Maybe it's just me, but I don't think of blocking / unblocking networking as Connect/Disconnect. Maybe Cordon/Uncordon or BlockNetwork/UnblockNetwork would be a bit better based on what I think these functions do.

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	c.Logger().Info(fmt.Sprintf("oldLeader=%v", oldLeader))
+
+	if err := oldLeader.Disconnect(ctx); err != nil {
+		return trace.Wrap(err, "failed to disconnect from cluster")
+	}
+
+	newLeader, err := getLeaderNode(ctx, nodes)
```
knisbet: This seems like it operates too quickly after Disconnect is called. Because the network is getting blocked, it's going to take the rest of the cluster some amount of time to realize the old leader has expired, and elect a new leader. So you'll need to wait/watch for the leadership change, and if it doesn't change after some timeout, then fail the test.

infra/gravity/node_commands.go
klizhentas: @knisbet @bernardjkim good point, in networking world the term is `Partition/Unpartition`

docker/Makefile
```diff
@@ -4,7 +4,7 @@ VERSION ?= $(shell git describe --long --tags --always|awk -F'[.-]' '{print $$1
 TAG ?= latest
 DOCKER_ARGS ?= --pull
 
-GRAVITY_VERSION := 5.2.3
+GRAVITY_VERSION := 5.2.15
```
r0mant: I would use a more recent gravity version, e.g. `6.1.x`.

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,118 @@
+package gravity
```
r0mant: License header is missing.

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"context"
+
+	"github.com/gravitational/robotest/lib/wait"
```
r0mant: Nit: We try to follow https://github.com/golang/go/wiki/CodeReviewComments#imports and thus organize all imports into 3 groups separated by an empty line: first stdlib, then imports from this project (`github.com/gravitational/robotest` in this case) and finally all other external imports.

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,118 @@
...
+	c.Logger().WithFields(logrus.Fields{
+		"oldLeader": oldLeader,
+	}).Info("Leader found")
+
+	if err := oldLeader.PartitionNetwork(ctx); err != nil {
+		return trace.Wrap(err, "failed to create network partition")
+	}
+	// NOTE: is subnets an appropriate name for the separate networks, or is the
+	// term only relevant when grouping a range of IP addresses?
+	subnets := make([][]Gravity, 0, 2)
```
r0mant: I think maybe "partitions" would be a bit better suitable name.

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return trace.Wrap(err, "new leader was not elected")
+	}
+
+	c.Logger().WithFields(logrus.Fields{
+		"oldLeader": oldLeader,
+		"newLeader": newLeader,
+	}).Info("New leader elected")
+
+	if err := oldLeader.UnpartitionNetwork(ctx); err != nil {
```
r0mant: Before restoring the network, should this run some tests on the second "partition" to make sure cluster is operational (e.g. gravity status, kubect get pods/nodes, etc.)?

infra/gravity/cluster_failover.go
```diff
@@ -0,0 +1,67 @@
...
+	if status.Cluster.Status != "active" {
+		return trace.Wrap(err)
+	}
+
+	return nil
+}
+
+// getLeaderNode returns the current leader node.
+func getLeaderNode(ctx context.Context, nodes []Gravity) (leader Gravity, err error) {
+	for _, node := range nodes {
```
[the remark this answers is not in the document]
r0mant: I think the retry is handled by the caller?

infra/gravity/node_commands.go
```diff
@@ -587,6 +604,66 @@ func (g *gravity) RunInPlanet(ctx context.Context, cmd string, args ...string) (
...
+	status, err := g.Status(ctx)
+	if err != nil {
+		return false
+	}
+	etcdLeaderKey := fmt.Sprintf("/planet/cluster/%s/master", status.Cluster.Cluster)
+
+	// TODO: use planet cmd once planet cli is updated s.t. `leader view` does
+	// not require args.
+	// leaderIP, err := g.RunInPlanet(ctx, "planet", "leader", "view")
+	leaderIP, err := g.RunInPlanet(ctx, "etcdctl", "get", etcdLeaderKey)
```
r0mant: Why not just use "planet leader view" with the same etcd key flag?

suite/sanity/failover.go
```diff
@@ -0,0 +1,37 @@
+package sanity
```
r0mant: Missing license header.

infra/gravity/node_commands.go
```diff
@@ -587,6 +604,66 @@ func (g *gravity) RunInPlanet(ctx context.Context, cmd string, args ...string) (
...
+// the cluster. If a network partition does not already exist, this will have
+// no effect.
+func (g *gravity) UnpartitionNetwork(ctx context.Context) error {
+	status, err := g.Status(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	for _, node := range status.Cluster.Nodes {
+		// TODO: look into using existing libs for manipulating network traffic.
+		cmdAcceptInput := fmt.Sprintf("iptables -I INPUT -s %s -j ACCEPT", node.Addr)
```
r0mant: I'm curious why you chose to insert another ACCEPT rule rather than deleting the DROP rule inserted by the partition method? Isn't the DROP one gonna be hit first?

infra/gravity/cluster_failover.go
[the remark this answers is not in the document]
bernardjkim: Oh, added retry after his comment, should have resolved convo.

infra/gravity/cluster_failover.go
bernardjkim: Should license header be included in every .go file?
Looks like all the files in in robotest are missing them.

r0mant: Yes, all source files should include the license header.

infra/gravity/node_commands.go
bernardjkim: Oh, that's probably a better way to handle it.

infra/gravity/node_commands.go
bernardjkim: Wasn't sure if the command worked properly. Was getting an error when I tested it on a cluster I had running locally. I'll retest it and replace it with proper planet command.

wadells:
@bernardjkim: What is the status of this?  Do you plan to ever move forward with this pr?  After #153 introduced go modules, this content got a bit stale, and is only getting more so.  While we definitely want to hang onto the branch, I'm unsure why we are keeping this PR open.

wadells:
Closing until such point as we pick this effort back up.

PR #174, opened 2020-02-07
Copyall fix

wadells:
As I've been messing with robotest, I found a bug with copyall.  It contains logic for the src being a regular file, but will fail to create a directory -- unlike when the source was a directory.

This level of testing is probably overblown in the long run, but because I'm still learning go, I'm using these early patches as a playground to work on my skills.  Any and all feedback is appreciated... I don't know much/anything about go standards and conventions yet.

Testing Done:

See the tests added.  Here's the run before & after:
```
walt@work:~/git/robotest$ git checkout HEAD^
# snip 'detached HEAD' message
HEAD is now at 305e10d Add tests for the copyAll function.
walt@work:~/git/robotest$ go test ./lib/system       
--- FAIL: TestCopyAllFileDstAbsent (0.00s)
    file_test.go:133: copy "/tmp/robotest-test138762263/src" -> "/tmp/robotest-test138762263/dst" failed: open /tmp/robotest-test138762263/dst/727014538: no such file or directory
    file_test.go:139: expected /tmp/robotest-test138762263/dst to be created
FAIL
FAIL    github.com/gravitational/robotest/lib/system    0.005s
FAIL
walt@work:~/git/robotest$ git checkout copyall-fix                                                                                                                                        [1]
Previous HEAD position was 305e10d Add tests for the copyAll function.
Switched to branch 'copyall-fix'
walt@work:~/git/robotest$ go test ./lib/system    
ok      github.com/gravitational/robotest/lib/system    (cached)
```

### DIFF
--- a/lib/system/file.go
+++ b/lib/system/file.go
@@ -77,27 +77,40 @@ func copyAll(src, dst string, fileCount *uint) (err error) {
 	dst = filepath.Clean(dst)
 
 	log.Debugf("copy %s -> %s", src, dst)
+
+	// fetch source permissions
 	si, err := os.Stat(src)
 	if err != nil {
 		return trace.ConvertSystemError(err)
 	}
+
+	// files often don't have execute, but directories need it
+	mode := si.Mode()
 	if si.Mode().IsRegular() {
-		*fileCount++
-		return CopyFile(src, filepath.Join(dst, filepath.Base(src)))
+		mode = mode | os.ModeDir
+		mode = mode | 0100
 	}
 
+	// ensure destination exists
 	_, err = os.Stat(dst)
 	if err != nil && !os.IsNotExist(err) {
 		return trace.ConvertSystemError(err)
 	}
 
 	if err != nil && os.IsNotExist(err) {
-		err = os.MkdirAll(dst, si.Mode())
+		err = os.MkdirAll(dst, mode)
 		if err != nil {
 			return trace.ConvertSystemError(err)
 		}
 	}
 
+	// if source is a file, copy
+	if si.Mode().IsRegular() {
+		*fileCount++
+		return CopyFile(src, filepath.Join(dst, filepath.Base(src)))
+	}
+
+	// if sorce is a directory, copy all contents
 	entries, err := ioutil.ReadDir(src)
 	if err != nil {
 		return trace.ConvertSystemError(err)
--- a/lib/system/file_test.go
+++ b/lib/system/file_test.go
@@ -77,20 +77,27 @@ func TestCopyAllDirDstAbsent(t *testing.T) {
 	}
 }
 
-func TestCopyAllFileDstPresent(t *testing.T) {
+// populates a test workspace (tmp) with a src file
+func createSourceFile(t *testing.T) (tmp, src string, fcount int, err error) {
 	// create a workspace
-	tmp, err := ioutil.TempDir("", "robotest-test")
+	tmp, err = ioutil.TempDir("", "robotest-test")
 	if err != nil {
 		t.Skipf("unable to create tempdir: %s", err)
 	}
-	defer os.RemoveAll(tmp)
 
 	// create source file
-	fcount := 1
-	src := filepath.Join(tmp, "/src")
+	fcount = 1
+	src = filepath.Join(tmp, "/src")
 	if err = ioutil.WriteFile(src, []byte("data"), 0640); err != nil {
 		t.Skipf("unable to write to %s: %s", src, err)
 	}
+
+	return tmp, src, fcount, nil
+}
+
+func TestCopyAllFileDstPresent(t *testing.T) {
+	tmp, src, fcount, err := createSourceFile(t)
+	defer os.RemoveAll(tmp)
 
 	// create destination directory
 	dst := filepath.Join(tmp, "/dst")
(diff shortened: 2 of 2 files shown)


lib/system/file_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+}
+
+func TestCopyAllFileDstAbsent(t *testing.T) {
+	// create a workspace
+	tmp, err := ioutil.TempDir("", "robotest-test")
+	if err != nil {
+		t.Skipf("unable to create tempdir: %s", err)
+	}
+	// defer os.RemoveAll(tmp)
```
a-palchikov: why is this commented out?

lib/system/file_test.go
```diff
@@ -0,0 +1,142 @@
...
+		t.Errorf("expected %v files to be copied, only %v copied", fcount, cnt)
+	}
+	if _, err := os.Stat(dst); os.IsNotExist(err) {
+		t.Errorf("expected %v to be created", dst)
+	}
+}
+
+func TestCopyAllFileDstPresent(t *testing.T) {
+	// create a workspace
+	tmp, err := ioutil.TempDir("", "robotest-test")
```
a-palchikov: nit: you have `createSourceDir` but no `createSourceFile` which needlessly complicates the file-specific tests.

lib/system/file.go
```diff
@@ -77,27 +77,40 @@ func copyAll(src, dst string, fileCount *uint) (err error) {
...
+	mode := si.Mode()
 	if si.Mode().IsRegular() {
-		*fileCount++
-		return CopyFile(src, filepath.Join(dst, filepath.Base(src)))
+		mode = mode | os.ModeDir
+		mode = mode | 0100
 	}
 
+	// ensure destination exists
 	_, err = os.Stat(dst)
```
bernardjkim: Is it necessary to check file info before attempting to make the directory? Will `Stat` return a different error than `MkdirAll`?

lib/system/file_test.go
wadells: Whoops!  That was left over from my test-testing. Will fix.

lib/system/file_test.go
wadells: My thought here was related to go's philosophy of "a little copying is better than a little dependency". 

https://go-proverbs.github.io/

When creating a directory with files in it, it was definitely enough work to abstract.  When creating a single file, it wasn't as clear to me that abstracting provided value, which is why I intentionally left it this way.

I trust your judgement though (since I'm learning what is appropriate in go) and will update!

lib/system/file_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// populates a test workspace (tmp) with a src containing several files
+func createSourceDir(t *testing.T) (tmp, src string, fcount int, err error) {
```
wadells: This function never returns an error.  In all error paths, it uses `t.Skipf` or some other form of abort.  Thus we can drop the extra error parameter.

lib/system/file_test.go
wadells: Addressed in https://github.com/gravitational/robotest/pull/174/commits/6e27e9e83a339f223c98b54e3660fa8b9fdef585

lib/system/file_test.go
wadells: Addressed in https://github.com/gravitational/robotest/pull/174/commits/676d981935804a3dd1bdc6265ab92afc7a3c6214

lib/system/file_test.go
wadells: Addressed in https://github.com/gravitational/robotest/pull/174/commits/66443c17c4e544f6433a34dc3acfe67927f343cc

wadells:
Tests still pass after updates & rebase:

```
walt@work:~/git/robotest$ go test ./lib/system    
ok      github.com/gravitational/robotest/lib/system    (cached)
```

lib/system/file_test.go
a-palchikov: I'm in favor of tests that I can read and immediately understand what's being tested and how. Here's the method I'm trying to stick to when writing tests: http://xunitpatterns.com/Four%20Phase%20Test.html
For me this means a test should focus as much as possible on a thing being tested and contain as little irrelevant or setup code as possible. Using the current implementation as an example:
```go
	// setup
	tmp, src, fcount := createSourceFile(t)
	defer os.RemoveAll(tmp)
	// no destination directory
	dst := filepath.Join(tmp, "/dst")

	// exercise
	var cnt uint
	err := copyAll(src, dst, &cnt)

	// validate
	if err != nil {
		t.Errorf("copy %q -> %q failed: %s", src, dst, err)
	}
	if int(cnt) != fcount {
		t.Errorf("expected %v files to be copied, only %v copied", fcount, cnt)
	}
	if _, err := os.Stat(dst); os.IsNotExist(err) {
		t.Errorf("expected %v to be created", dst)
	}
```
Has nothing that would detract from understanding what the test case is about.

wadells:
> Also, we seem to have pulled at least a couple of higher-level testing libraries (ginko and testify) and the internal tests (not the robotest implementation) seems to be a split between built-in tests and testify. It would be nice if we could stick to a single library.

I've only used go's built in tests thus far.  I'll take a look at ginko and testify, and how they're currently used around Gravitational.  Do you have a preference @a-palchikov, or is this something you'd like to delegate?

FWIW, I've used a lot of different test harnesses & frameworks in other languages, so I'm comfortable assessing the utility of them.  I haven't yet found the priority to explore the variety of testing available in golang yet.  My learning philosophy is that is best to start with first principles (the built-in `go test` stuff) and build my knowledge out from there.

lib/system/file_test.go
wadells: Thanks for the link & the snippet.  I agree with your remark that

> a test should focus as much as possible on a thing being tested and contain as little irrelevant or setup code as possible

I typically lean towards fixtures for setup/teardown--to keep that out of the test case itself. pytest's parameterized fixtures are probably my favorite implementation (if a bit _magic_).  But I haven't reached that level of sophistication in go yet, as these are the first tests I've contributed in the language.  In lieu of fixtures, abstracting out common setup logic into a function does well enough.  Thanks for suggesting it.

PR #185, opened 2020-03-12
A couple of tweaks

a-palchikov:
* Moves status query before the call to `uploadUpgrade` to counter recent test failures. `uploadUpgrade` only verifies the status in the cluster record and might fail due to an intermittent issue. This is not an ideal solution but might help avoid test failures.
 * Removes state/etcd directories from the chowned list - these will be configured by the installer. The reason this is not done in the bootstrap script is that the script also runs on reboot and will break
the access to directories that aren't expected to be owned by the service user (which is especially critical with SELinux enabled).

### DIFF
--- a/infra/gravity/cluster_install.go
+++ b/infra/gravity/cluster_install.go
@@ -190,10 +190,6 @@ func (c *TestContext) Upgrade(nodes []Gravity, installerURL, gravityURL, subdir 
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = c.Status(nodes)
-	if err != nil {
-		return trace.Wrap(err)
-	}
 	return c.upgrade(master, len(nodes))
 }
 
@@ -205,6 +201,11 @@ func (c *TestContext) uploadInstaller(master Gravity, nodes []Gravity, installer
 	defer cancel()
 
 	err := master.SetInstaller(ctx, installerURL, subdir)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	err = c.Status(nodes)
 	if err != nil {
 		return trace.Wrap(err)
 	}
--- a/assets/terraform/gce/bootstrap/centos.sh
+++ b/assets/terraform/gce/bootstrap/centos.sh
@@ -54,7 +54,7 @@ function setup-user {
     retry restorecon -vR /home/${os_user}
   fi
 
-  chown -R $service_uid:$service_gid /var/lib/gravity $etcd_dir /home/${os_user}
+  chown -R $service_uid:$service_gid /home/${os_user}
   sed -i.bak 's/Defaults    requiretty/#Defaults    requiretty/g' /etc/sudoers
 }
 
--- a/assets/terraform/gce/bootstrap/suse.sh
+++ b/assets/terraform/gce/bootstrap/suse.sh
@@ -41,7 +41,7 @@ function setup-user {
     chmod 0600 /home/${os_user}/.ssh/authorized_keys
   fi
 
-  chown -R $service_uid:$service_gid /var/lib/gravity $etcd_dir /home/${os_user}
+  chown -R $service_uid:$service_gid /home/${os_user}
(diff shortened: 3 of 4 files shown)


a-palchikov:
I don't have failed robotest run for the `chown` issue as this is not something that'd happen during such a rub. But the I do use the terraform scripts to boot VMs manually that I'll be stopping/starting multiple times - this is when it becomes an issue.

The upgrade issue is visible [here](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&angularJsUrl=%2Flogs%2Fviewer%3FadvancedFilter%3Dresource.type%253D%2522project%2522%250Alabels.__uuid__%253D%2522a3a287db-4226-47d4-a0a3-c78ace0803da%2522%250Alabels.__suite__%253D%252271b3df61-8fd0-4d11-a36b-efec5adba9fc%2522%250Aseverity%253E%253DINFO%26authuser%3D1%26expandAll%3Dfalse%26project%3Dkubeadm-167321&minLogLevel=0&timestamp=2020-03-13T15:24:41.055000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22a3a287db-4226-47d4-a0a3-c78ace0803da%22%0Alabels.__suite__%3D%2271b3df61-8fd0-4d11-a36b-efec5adba9fc%22%0AjsonPayload.message:%22cluster%20is%20in%20degraded%22&interval=NO_LIMIT&scrollTimestamp=2020-03-11T18:00:30.810644239Z)

wadells:
> I don't have examples of robotest runs for the `chown` issue as this is not something that'd happen during such a run. But I do use terraform scripts to boot VMs manually that I'll be stopping/starting multiple times - this is when it becomes an issue.

Thanks for explaining.  I haven't used this workflow yet, but I have considered it as a  "hack" for trying to repro issues.  I'll keep it in mind as a secondary use case for future work.

> The upgrade issue is visible [here](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&angularJsUrl=%2Flogs%2Fviewer%3FadvancedFilter%3Dresource.type%253D%2522project%2522%250Alabels.__uuid__%253D%2522a3a287db-4226-47d4-a0a3-c78ace0803da%2522%250Alabels.__suite__%253D%252271b3df61-8fd0-4d11-a36b-efec5adba9fc%2522%250Aseverity%253E%253DINFO%26authuser%3D1%26expandAll%3Dfalse%26project%3Dkubeadm-167321&minLogLevel=0&timestamp=2020-03-13T15:24:41.055000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22a3a287db-4226-47d4-a0a3-c78ace0803da%22%0Alabels.__suite__%3D%2271b3df61-8fd0-4d11-a36b-efec5adba9fc%22%0AjsonPayload.message:%22cluster%20is%20in%20degraded%22&interval=NO_LIMIT&scrollTimestamp=2020-03-11T18:00:30.810644239Z).

Thanks!

infra/gravity/cluster_install.go
```diff
@@ -190,10 +190,6 @@ func (c *TestContext) Upgrade(nodes []Gravity, installerURL, gravityURL, subdir
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = c.Status(nodes)
```
r0mant: I wonder why didn't you just move it before uploadInstalled above, but moved it into the upload function.

wadells:
@a-palchikov Are we good to merge here?  I'd like to get this PR closed out. @r0mant's remark seems more like a "why?" than a bug or suggestion, so it probably isn't a blocker.

infra/gravity/cluster_install.go
a-palchikov: I moved closer to the actual `upload upgrade` call - otherwise it would be two `Status` calls one after the other.

Issue #201, opened 2020-04-01
Add the ability to test SELinux


wadells:
With @a-palchikov's recent work in https://github.com/gravitational/gravity/pull/1108 https://github.com/gravitational/gravity/pull/1172, Gravity now has the ability to run on distros hardened by SELinux.

Robotest should offer consumers (gravity CI mostly) a way to exercise gravity on a SELinux hardened distro.

a-palchikov:
This can be configured in a noninvasive way by setting the [environment variable](https://github.com/gravitational/gravity/blob/7.0.0-rc.5/lib/defaults/defaults.go#L868) `GRAVITY_SELINUX=true` for the gravity install process.

wadells:
> This can be configured in a noninvasive way by setting the [environment variable](https://github.com/gravitational/gravity/blob/7.0.0-rc.5/lib/defaults/defaults.go#L868) `GRAVITY_SELINUX=true` for the gravity install process.

@a-palchikov Can you elaborate a bit more?

Upon preliminary investigation, neither install nor join has environment variables plumbed through:

https://github.com/gravitational/robotest/blob/95865405776d1185dcb94038ceb7ae65ec0aaed9/infra/gravity/node_commands.go#L263
https://github.com/gravitational/robotest/blob/95865405776d1185dcb94038ceb7ae65ec0aaed9/infra/gravity/node_commands.go#L354
https://github.com/gravitational/robotest/blob/95865405776d1185dcb94038ceb7ae65ec0aaed9/lib/ssh/run_parse.go#L21

So I don't think this will be particularly non-invasive.  Unless you're suggesting we set this environment variable elsewhere (e.g. the node's shell profile).

For consistency's sake, I'd suggest we go with gravity CLI flags and the `InstallParam` struct, as this is how most (all?) install configuration is handled

https://github.com/gravitational/robotest/blob/95865405776d1185dcb94038ceb7ae65ec0aaed9/infra/gravity/node_commands.go#L267-L280

a-palchikov:
By non-invasive I meant for a particular installer - not the robotest binary which will need to set corresponding environment variables (perhaps only for relevant distributions).

wadells:
It appears that the current robtest selinux triggerst don't actually enable gravity's SELinux support.  Reopening.

Issue #182, opened 2020-03-07
Fix Debian 8 support


wadells:
**Summary**
Robotest needs to support RHEL 7 testing for Gravity 5.5.x+

**Context**
We have important customers gravity in the field running 5.5.x on RHEL.  @eldios mentioned:

> RHEL or CentOS usually latest version available or the one prior (in this case 7 or 8 )

RHEL 7 is supported in Gravity 5.0.0+

https://github.com/gravitational/gravity.e/blob/94ebe01f6d689b7bd37db0b94b960b3235e182fd/assets/telekube/resources/app.yaml#L77-L78

However in Gravity 5.5.x CI job, we're seeing install failures on a variety of 5.X upgrade scenarios including "redhat:7" and "debian:8", after I attempted to introduce these in https://github.com/gravitational/gravity/commit/5d3f2941c6666252ed3ed2cddafe30895e3a269f

Here's a 3 node install 5.5.38-dev.X x redhat:7 test that fails: [stackdriver logs](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-03-06T17:32:49.623000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22df0a0f46-5f3c-4812-9015-18e35ca33d54%22%0Alabels.__suite__%3D%226ccb358e-7150-4742-a83c-22b79c5b7d57%22%0Aseverity%3E%3DINFO&dateRangeEnd=2020-03-06T17:32:47.905Z&interval=PT1H&dateRangeUnbound=backwardInTime&scrollTimestamp=2020-03-06T04:35:43.450935185Z)

[...]

PR #205, opened 2020-04-14
Add SELinux triggers for SELinux-enabled hosts

a-palchikov:
Fixes https://github.com/gravitational/robotest/issues/201.

I will validate the change manually before merging.

### DIFF
--- a/infra/gravity/node_commands.go
+++ b/infra/gravity/node_commands.go
@@ -260,7 +260,11 @@ func (g *gravity) Install(ctx context.Context, param InstallParam) error {
 		return trace.Wrap(err, buf.String())
 	}
 
-	err = sshutils.Run(ctx, g.Client(), g.Logger(), buf.String(), nil)
+	err = sshutils.Run(ctx, g.Client(), g.Logger(), buf.String(),
+		// Add SELinux-specific triggers. These will only have effect on SELinux-enabled hosts
+		map[string]string{
+			constants.GravitySELinuxEnv: "true",
+		})
 	return trace.Wrap(err, param)
 }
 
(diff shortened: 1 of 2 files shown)


a-palchikov:
I'm not sure whether there are `correct` images w.r.t SELinux - `rhel` and `CentOS` have SELinux turned on by default - as if it's configuration by convention. I would be surprised to have a corresponding OS distro (from the two mentioned) image where it was disabled by default.

wadells:
> I'm not sure whether there are `correct` images w.r.t SELinux - `rhel` and `CentOS` have SELinux turned on by default - as if it's configuration by convention. I would be surprised to have a corresponding OS distro (from the two mentioned) image where it was disabled by default.

I think I I understand now. This change will turn all of our RHEL (we don't run any at the moment because of #182)  and CentOS (which is regularly tested) test runs into SELinux test runs.

Do we care about testing RHEL/CentOs without Gravity's SELinux support?  E.g is that an important use case to customers in the field and/or have interesting different code paths?

Effectively, I'm trying to figure out if using SELinux should be configurable, and if so how that configuration is provided to Robotest consumers.  If you don't think it needs to be configurable (a.k.a Gravity on CentOS/RHEL without SELinux isn't used or isn't likely to regress), that makes life easy, and there is no more work to do in Robotest.

wadells:
> I'm not sure whether there are `correct` images w.r.t SELinux - `rhel` and `CentOS` have SELinux turned on by default - as if it's configuration by convention. I would be surprised to have a corresponding OS distro (from the two mentioned) image where it was disabled by default.

I guess I'm a bit confused then.  AFAIK we didn't support SELinux in Gravity until recently, ergo the centos we've been testing against must have had it disabled.

wadells:
Sorry about the accidental close. :smile:  I was trying to discard an earlier draft of my comment above.

a-palchikov:
I think it would be beneficial to test both sides on rhel or CentOS (does not have to be both distributions for each configuration though).

PR #208, opened 2020-04-22
Update images on GCE

a-palchikov:
Update images on GCE. Also added `rhel:8` and ``centos:8`.

I've removed `devicemapper`/`lvm` packages as we're not testing docker's `devicemapper` storage driver anymore.


assets/terraform/gce/os.tf
```diff
@@ -10,17 +10,20 @@ variable "oss" {
...
-    "debian:8"      = "debian-cloud/debian-8-jessie-v20180401"
-    "debian:9"      = "debian-cloud/debian-9-stretch-v20190813"
-    "debian:latest" = "debian-cloud/debian-9-stretch-v20190813"
-    "suse:12"       = "suse-cloud/sles-12-sp4-v20190617"
-    "suse:15"       = "suse-cloud/sles-15-sp1-v20190828"
-    "suse:latest"   = "suse-cloud/sles-15-sp1-v20190828"
+    "ubuntu:16"     = "ubuntu-os-cloud/ubuntu-1604-xenial-v20200407"
+    "ubuntu:18"     = "ubuntu-os-cloud/ubuntu-1804-bionic-v20200414"
+    "ubuntu:19"     = "ubuntu-os-cloud/ubuntu-1904-eoan-v20200413a"
+    "ubuntu:latest" = "ubuntu-os-cloud/ubuntu-1904-eoan-v20200413a"
```
wadells, line 16: I don't believe Gravity officially supports 19.04: https://gravitational.com/gravity/docs/requirements/.  Do we have a reason to test on it (e.g. a customer using it)?

Since focal released today, I'd swap this out to 20.04, which we will support as the next Ubuntu LTS.

assets/terraform/gce/os.tf
```diff
@@ -10,17 +10,20 @@ variable "oss" {
...
+    "redhat:7"      = "rhel-cloud/rhel-7-v20200420"
+    "redhat:8"      = "rhel-cloud/rhel-8-v20200413"
+    "centos:7"      = "centos-cloud/centos-7-v20200420"
+    "centos:8"      = "centos-cloud/centos-8-v20200413"
+    "debian:9"      = "debian-cloud/debian-9-stretch-v20200420"
```
wadells, line 21: We say Gravity supports Debian 8 here:  https://gravitational.com/gravity/docs/requirements/  What is your thinking behind removing it?  I recognize we don't run those tests presently, and I'd like to figure out if we should either:

1) plan to test it (https://github.com/gravitational/robotest/issues/182)
2) remove it from the list of distros that Gravity supports (would that be an 8.0 thing given semver?)

a-palchikov: Yes, I fell into a trap of thinking google is no longer supporting the `jessie`, but it does show up under deprecated. I'll put it back in.

Issue #171, opened 2020-01-24
Robotest needs its own CI


wadells:
As of 2020-01, there is no build/sanity check/static analysis run on each pull requests to the gravitational/robotest repo.  This means that issues like #170 can slip into the repo unnoticed.

We should at least run the container build and test make targets on every pull request.  More can be added over time.

klizhentas:
Please use force.gravitational.io - it's capable of doing basic CI, and I need more users

PR #209, opened 2020-05-05
Pass `os` and `nodes` via robotest.tfvars.json.

wadells:
## Description
This is a maintainer Quality of Life change.

These two parameters were previously handled as cli flags. This meant we
had two paths for routing data, and more annoyingly: it wasn't easy to
recreate a robotest `terraform apply` or `terraform destroy` via the
command line because these two values needed to be manually specified in
addition to the other values that came from robotest.tfvars.json.

## Type of change
* Internal change (not necessarily or a new feature)

## Linked tickets and other PRs
N/A.  

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Write documentation
- [x] Address review feedback

## Testing done
When I have a development run leak resources, and I try to manually clean up, I will navigate to the terraform directory and run `terraform destroy`.  This previously resulted in the following prompt:

```
walt@work:~/git/robotest/logs/0504-1616/walt/install-1/ubuntu18/overlay2/1n/tf$ terraform destroy -var-file robotest.tfvars.json
var.os
  Linux distribution as name:version, i.e. debian:9

  Enter a value: 
```

And after these changes:

```
walt@work:~/git/robotest/logs/0504-2221/walt/install-1/ubuntu18/overlay2/1n/tf$ cat robotest.tfvars.json 
{
  "credentials": "/home/walt/git/robotest/kubeadm-167321-2d663815917d.json",
  "node_tag": "robotest-a56ae835",
  "nodes": 1,
  "os": "ubuntu:18",
  "os_user": "ubuntu",
  "region": "us-west1",
  "ssh_pub_key_path": "/home/walt/.ssh/robotest.pub",
  "vm_type": "custom-8-8192"
 }
walt@work:~/git/robotest/logs/0504-2221/walt/install-1/ubuntu18/overlay2/1n/tf$ terraform destroy -var-file robotest.tfvars.json 
data.template_file.bootstrap: Refreshing state...
data.google_compute_zones.available: Refreshing state...
data.google_compute_subnetwork.robotest: Refreshing state...
data.google_compute_network.robotest: Refreshing state...
// snip
```

No os prompt, correct number of nodes for the run present in the vars file.

### DIFF
--- a/infra/terraform/terraform.go
+++ b/infra/terraform/terraform.go
@@ -213,8 +213,6 @@ func (r *terraform) Destroy(ctx context.Context) error {
 	varsPath := filepath.Join(r.stateDir, tfVarsFile)
 	destroyCommand := []string{
 		"destroy", "-auto-approve",
-		"-var", fmt.Sprintf("nodes=%d", r.NumNodes),
-		"-var", fmt.Sprintf("os=%s", r.OS),
 		fmt.Sprintf("-var-file=%s", varsPath),
 	}
 	if r.VarFilePath != "" {
@@ -306,15 +304,13 @@ func (r *terraform) boot(ctx context.Context) (rc io.ReadCloser, err error) {
 	}
 
 	varsPath := filepath.Join(r.stateDir, tfVarsFile)
-	err = r.saveVarsJSON(varsPath)
+	err = r.saveTerraformVars(varsPath)
 	if err != nil {
 		return nil, trace.Wrap(err, "failed to store terraform vars")
 	}
 
 	applyCommand := []string{
 		"apply", "-input=false", "-auto-approve",
-		"-var", fmt.Sprintf("nodes=%d", r.NumNodes),
-		"-var", fmt.Sprintf("os=%s", r.OS),
 		fmt.Sprintf("-var-file=%s", varsPath),
 	}
 	if r.VarFilePath != "" {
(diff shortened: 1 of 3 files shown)


infra/terraform/terraform.go
```diff
@@ -379,31 +375,61 @@ func (r *terraform) command(ctx context.Context, args []string, opts ...system.C
...
+	tfvars = make(map[string]interface{})
+	switch cfg.CloudProvider {
 	case constants.AWS:
-		config = r.Config.AWS
+		return nil, trace.NotImplemented("aws robotest hasn't been run in a while")
 	case constants.Azure:
-		config = r.Config.Azure
-	case constants.GCE:
-		config = r.Config.GCE
+		return nil, trace.NotImplemented("azure robotest hasn't been run in a while")
```
wadells, line 385: I don't like that I explicitly break aws and azure here, but I haven't yet found the priority to test on either of those platforms, and I'm not aware of anyone who currently uses them.

I can come pencil in some (probably untested) code here if you think it is worthwhile.

infra/terraform/terraform_test.go
```diff
@@ -0,0 +1,97 @@
+package terraform
+
+import (
+	"encoding/json"
+	"github.com/gravitational/robotest/infra/providers/gce"
```
a-palchikov, line 5: nit: this should be in a separate import group

infra/terraform/terraform_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestConvertConfigToTerraformVars(t *testing.T) {
+	gce_config := gce.Config{
```
a-palchikov, line 11: `gceConfig`

infra/terraform/terraform_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	b, err := json.Marshal(configMap)
+	e, err := json.Marshal(expected)
+	if !reflect.DeepEqual(b, e) {
+		t.Errorf("\ngot:\t\t%q\nexpected:\t%q", b, e)
+	}
+}
+
+func TestConvertConfigToTerraformVarsOptionalValues(t *testing.T) {
+	gce_config := gce.Config{
```
a-palchikov, line 52: gceConfig

infra/terraform/terraform.go
a-palchikov: No, it's fine.

infra/terraform/terraform_test.go
wadells: fixed in 8eabb37

infra/terraform/terraform_test.go
wadells: fixed in 8eabb37

infra/terraform/terraform_test.go
wadells: fixed in 8eabb37.  Down the road I'd like to see if I can implement this as a lint rule/guideline.  I'm kinda surprised gofmt/goimports doesn't automatically take care of this.

a-palchikov: It's a good question - which one are you using and how (i.e. directly or with editor/IDE integration)?

wadells: I use vim-go, which I just learned runs at 'vanilla' gofmt upon buffer save.  I checked `goimports` now, and it fixes this issue.  I'll fix my vimrc to change the default to goimports. :facepalm: 

Regardless of personal editor setups, I'd like our `make lint` target to catch nits like this. Once https://github.com/gravitational/robotest/issues/171 hits then we can largely save human reviewers the time, and encourage folks like me to configure their editors the right way.

I'll look at what it would take to do that.

Issue #200, opened 2020-04-01
Standardize Robotest's release process


wadells:
We should:

* return to semver labeling of releases, including following their breaking change & feature guidelines
* ensure all official robotest releases are tagged in git (ideally with signed tags)
* ensure all official robotest release have a GitHub release

### Context
As of 2020-04 last several robotest quay.io container tags used in Gravity CI are:

* dima
* stable-gce
* uid-gid

There has not been a release tagged in git or GitHub since 1.0 in April 2017 (three years ago at the time of filing).

Furthermore, there is not a clear way to go from quay.io robotest container to the hash of the git commit used to create that container, so tags like 'uid-gid' really are floating off in space, with no easy context as to their origin.

The current strategy makes it difficult for consumers of robotest to tell what the ordering of releases is, or if these releases are official (vs ad-hoc tags from some developers branch)

wadells:
I cut 2.0.0-rc.1 complete with signed tags and a github release.  I'm testing it out in https://github.com/gravitational/gravity/pull/1646.  Assuming no major issues are found in this version (which has all of the recent build/release changes), the only remaining work for this ticket is documenting Robotest's release process.

PR #215, opened 2020-05-08
Add version & commit information to logs.

wadells:
With #206 merged and version/commit information available within the go runtime, I now want to make sure that this info is recorded & accessible for all robotest runs.

Immediately, this is motivated by trying & failing to RCA https://github.com/gravitational/robotest/issues/214#issuecomment-625551626

I'd love to be able to make sure I'm running the exact same version.  I recognize this pach won't help with #214 because the cause predates all the version work.  But maybe it'll help me next time.

## Testing Done
Via the console:

```
+ dlv test ./suite -- -test.timeout=1h -gcl-project-id=<redacted> -test.parallel=1 -repeat=1 -fail-fast=true -always-collect-logs=true -resourcegroup-file=/home/walt/git/robotest/logs/0507-1831/alloc.txt -destroy-on-success=true -destroy-on-failure=true -tag=walt -suite=sanity -debug '-provision=
installer_url: s3://hub.gravitational.io/gravity/oss/app/telekube/7.0.4/linux/x86_64/telekube-7.0.4-linux-x86_64.tar
gravity_url: /home/walt/go/bin/gravity
script_path: /home/walt/git/robotest/assets/terraform/gce
tf_plugin_dir: /home/walt/.terraform.d/plugins/linux_amd64
state_dir: /home/walt/git/robotest/logs/0507-1831
cloud: gce
aws:
  access_key: <redacted>
  secret_key: <redacted>
  region: us-east-1
  ssh_user: unused
  key_path: unused
  key_pair: unused
  docker_device: unused
  vpc: unused
gce:
  credentials: <redacted>
  vm_type: custom-8-8192
  region: northamerica-northeast1,us-west1,us-west2,us-east1,us-east4,us-central1
  ssh_key_path: /home/walt/.ssh/robotest
  ssh_pub_key_path: /home/walt/.ssh/robotest.pub
  var_file_path: /home/walt/git/robotest/logs/0507-1831/gcp-vars.json
' 'noop={"nodes":1,"flavor":"one","os":"ubuntu:18","storage_driver":"overlay2","role":"node"}'
Type 'help' for list of commands.
(dlv) c
INFO[0000] [PROFILING] http localhost:6060              
DEBU[0000] Version:	1.0.1-dev.235+725c978c              
DEBU[0000] Git Commit:	725c978c75b1883fa96c860851c595d0ab1bf424
[...]

PR #206, opened 2020-04-15
Makefile work (version improvments, help target)

wadells:
In preparation for Robotest 2.0 (nothing fancy, it is a formality to acknowledge previous breaking changes) and to address #200 this changeset refactors the Robotest Makefile to have robust semver compliant versioning and docker tagging.  `make help` thrown in as a bonus.

I explored using the following solutions, in addition to the one proposed here:

 1. https://github.com/gravitational/version Test branch here: https://github.com/wadells/robotest/commit/97b20ed0d17d78e51d91de15753d3c0b38163481  I dismissed this because I found the linkflags passing of information and extra dependecy to be obtuse compared to a generated `version.go`.

2. https://github.com/mdomke/git-semver This repo has clearly grappled with the exact same problems that motivated my changes.  I still found that <40 lines of code in make/shell to be prefereable to a new tool in the toolchain that isn't used elsewhere in gravitational.

### Warning!
This will cause our semver version to regress from `1.0.2XX` to `1.0.1-dev.2.XX`  However, I plan to create a 2.0 alpha tag as soon as this is in.  As for 1.0.X: There have already been many breaking changes in 1.0.X patch releases, so the 1.0.X line of development is not properly semver compliant anyhow.  I do not plan to continue to support 1.0.X once 2.0.0 is out.

### Testing Done:
<details>
<summary><code>make help</code> testing</summary>
<pre>
walt@work:~/git/robotest$ make help
Usage: make [options] [target] ...

Options: run `make --help` for options

Targets:
  all             Clean and build.
  build           Compile go binaries.
  clean           Remove intermediate build artifacts & cache.
  containers      Build container images.
  help            Show this message.
  lint            Run static analysis against source code.
  publish         Publish container images to quay.io.
  test            Run unit tests.
  version         Show the robotest version.
</pre>
</details>

<details>
<summary><code>make version</code> testing</summary>
<pre>
walt@work:~/git/robotest$ make version
version metadata saved to version.go
Robotest Version: 1.0.1-dev.221+d40bd462
walt@work:~/git/robotest$ git tag 2.0.0-alpha
walt@work:~/git/robotest$ make version       
version metadata saved to version.go
Robotest Version: 2.0.0-alpha
walt@work:~/git/robotest$ git tag -d 2.0.0-alpha
Deleted tag '2.0.0-alpha' (was d40bd46)
walt@work:~/git/robotest$ git tag 2.0.0-alpha HEAD~5
walt@work:~/git/robotest$ make version
version metadata saved to version.go
[...]

PR #221, opened 2020-05-13
Strip leading 'v' from git tags if present.

wadells:
As discussed at:

  https://github.com/golang/go/issues/32945
  https://semver.org/#is-v123-a-semantic-version

Go and semver disagree about handling of a leading 'v' on version
strings.  This patch allows robotest to play nicely with both.

I discovered this when I created the v2.0.0-alpha.1 tag in git.

Contributes to https://github.com/gravitational/robotest/issues/200.

### Testing Done
Before:
```
walt@work:~/git/robotest$ git describe --tags
v2.0.0-alpha.1
walt@work:~/git/robotest$ make version       
version metadata saved to version.go
Robotest Version: v2.0.0-alpha.1
```
After:
```
walt@work:~/git/robotest$ git describe --tags        
v2.0.0-alpha.1-1-g4ac1444
walt@work:~/git/robotest$ make version       
version metadata saved to version.go
Robotest Version: 2.0.0-alpha.1.1+4ac14441
```
With a non-v-prefixed tag:
```
walt@work:~/git/robotest$ git tag 3.fake.version
walt@work:~/git/robotest$ make version
version metadata saved to version.go
Robotest Version: 3.fake.version
walt@work:~/git/robotest$ git tag -d 3.fake.version 
Deleted tag '3.fake.version' (was 4ac1444)
```

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
 # this versioning algo:
-#  - if on a tagged commit, use only the tag
-#    e.g. 6.2.18 (for the commit tagged 6.2.18)
+#  - if on a tagged commit, use the tag with any preceding 'v' stripped
+#    e.g. 6.2.18 (for the commit tagged v6.2.18)
 #  - if last tag was a regular release, bump the minor version, make a it a 'dev' pre-release, and append # of commits since tag
-#    e.g. 5.5.38-dev.5 (for 5 commits after 5.5.37)
+#    e.g. 5.5.38-dev.5 (for 5 commits after v5.5.37)
 #  - if last tag was a pre-release tag (e.g. alpha, beta, rc), append number of commits since the tag
-#    e.g. 7.0.0-alpha.1.5 (for 5 commits after 7.0.0-alpha.1)
+#    e.g. 7.0.0-alpha.1.5 (for 5 commits after v7.0.0-alpha.1)
 
 
 increment_patch() {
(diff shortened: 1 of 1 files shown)


version.sh
```diff
@@ -29,11 +29,20 @@ COMMITS_SINCE_LAST_TAG=`git rev-list  ${COMMIT_WITH_LAST_TAG}..HEAD --count`
 BUILD_METADATA=`git rev-parse --short=8 HEAD`
 DIRTY_AFFIX=$(git diff --quiet || echo '-dirty')
 
+# strip leading v from git tag, see:
```
a-palchikov, line 32: Header comment should be updated to reflect the change.
On a side note - what benefit does this bring vs tagging without the `v`?

wadells: >  On a side note - what benefit does this bring vs tagging without the v?

TLDR: `v` prefixes are what the wider go ecosystem use

To quote the go developers:

> The difference is mainly aesthetic and relatively arbitrary. Absent a compelling need otherwise, Go usually resolves aesthetic differences by picking one to use consistently (think of gofmt), not supporting both.

https://github.com/golang/go/issues/32945#issuecomment-512900259

`v` prefixed git tags are the right choice for go libraries because `go mod` reverts to pseudo-versions if there is no `v` in the git tag.  And if we're going to track libraries via a pseudo version, why bother to make git version tags at all?

A quite reasonable counter argument is: YAGNI, robotest will never be used as a library.  However why have  different version tagging strategies for libraries vs applications? "Resolve aesthetic differences by picking one to use consistently"

I know Gravity made a choice to use non-v-prefixed before go modules were widely available, but I doubt we'd make the same choice today for a new go project.  You could argue robotest made the same choice, but two tags (0.0.1 and 1.0.0 -- neither of which is used today) don't carry the same responsibility to the customers of the project as Gravity's longstanding & widely used release tags.

Effectively, I'm treating 2.0.0 as the first semantically versioned release, and everything prior as "dragons be here" jumbled prehistory  -- in a large part because it isn't easy determine which robotest code is running before the recent versioning logging changes (https://github.com/gravitational/robotest/pull/206 and https://github.com/gravitational/robotest/pull/215) that will be available in 2.0.0.

/ramble

wadells: I updated the comments in fb8125a.  Thanks for helping me keep up with documentation.

PR #229, opened 2020-05-20
Add a shrink test scenario.

### LINKED ISSUE #188: Add cluster shrink test scenario
Shrinks are common in the field, esp for critical customers with autoscaling groups.  We should have a shrink specific test.

wadells:
Fixes #188.

I also refactored RemoveNode -> RemoveNodes for signature consistency
with Expand() and the unused ShrinkLeave().

### Testing After PR Feedback
5.5.40 [shrink logs](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-22T02:37:23.621000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%225b9df45e-f63f-47f5-babb-2f881dfb4b1f%22%0Alabels.__suite__%3D%227b42a0e3-9bce-42ce-82b1-c608492facb3%22%0Aseverity%3E%3DINFO&dateRangeStart=2020-05-22T01:37:24.121Z&dateRangeEnd=2020-05-22T02:37:24.121Z&interval=PT1H&scrollTimestamp=2020-05-22T02:37:04.145726434Z)

### Testing Done
7.0.5 [logs](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-20T05:02:13.690000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22e1dbcb40-c486-40a5-b294-4f8578d0db80%22%0Alabels.__suite__%3D%22188d1189-da74-47d8-89c0-3a4ed268ea08%22%0Aseverity%3E%3DINFO&dateRangeStart=2020-05-20T03:39:51.336Z&interval=PT1H&scrollTimestamp=2020-05-20T04:47:53.927326441Z&dateRangeUnbound=forwardInTime)

Unfortunately, reproing https://github.com/gravitational/gravity/issues/1445 is flaky.  Here are two example runs:

5.5.40 [repro](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-20T04:40:22.738000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%220081a1c1-d0e5-459b-ae00-161b6700cb82%22%0Alabels.__suite__%3D%2276ae03e0-d6fa-4bf8-a250-fd065625d691%22&dateRangeStart=2020-05-20T03:42:14.800Z&dateRangeEnd=2020-05-20T04:42:14.800Z&interval=PT1H&scrollTimestamp=2020-05-20T04:09:28.820653764Z) (there was an unrelated failure to log/cleanup on this run)
5.5.40 [no repro](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-20T04:33:18.953000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%226fb81578-2588-462e-8b00-1bec322971d1%22%0Alabels.__suite__%3D%22bdea6413-a8ce-4216-8fcf-e29eae77bb0c%22%0Aseverity%3E%3DINFO&dateRangeStart=2020-05-20T03:33:19.512Z&dateRangeEnd=2020-05-20T04:33:19.512Z&interval=PT1H&scrollTimestamp=2020-05-20T04:32:31.947230235Z)

As I've already invested the better part of a week into manually reproing, and attempting to debug, I'm calling this good enough for now.  Let me know if you have ideas about how we could make this test (or the product) more deterministic.

### DIFF
--- a/infra/gravity/cluster_resize.go
+++ b/infra/gravity/cluster_resize.go
@@ -3,47 +3,33 @@ package gravity
 import (
 	"context"
 
-	"github.com/gravitational/robotest/lib/constants"
-	"github.com/gravitational/robotest/lib/utils"
-
 	"github.com/gravitational/trace"
-	"github.com/sirupsen/logrus"
 )
 
-func (c *TestContext) Expand(current, extra []Gravity, p InstallParam) error {
-	if len(current) == 0 || len(extra) == 0 {
+// Expand joins one or more node to a cluster
+func (c *TestContext) Expand(currentCluster, nodesToJoin []Gravity, p InstallParam) error {
+	if len(currentCluster) < 1 {
 		return trace.BadParameter("empty node list")
 	}
-	if c.provisionerCfg.CloudProvider == constants.Ops {
-		return trace.NotImplemented("not implemented")
+	if len(nodesToJoin) < 1 { // nothing to be done
+		return nil
 	}
 
-	c.Logger().WithFields(logrus.Fields{
-		"current": current,
-		"extra":   extra,
-	}).Info("Expand.")
-
+	// status is solely used for gathering the join token, can this be replaced
+	// with InstallParam.Token -- 2020-05 walt
+	peer := currentCluster[0]
 	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
 	defer cancel()
-
-	master := current[0]
-	joinAddr := master.Node().PrivateAddr()
-	status, err := master.Status(ctx)
+	status, err := peer.Status(ctx)
 	if err != nil {
-		return trace.Wrap(err, "query status from [%v]", master)
+		return trace.Wrap(err, "Query status from [%v]: %v", peer, err)
 	}
+	token := status.Cluster.Token.Token
 
-	ctx, cancel = context.WithTimeout(c.ctx, withDuration(c.timeouts.Install, len(extra)))
-	defer cancel()
+	c.Logger().WithField("current", currentCluster).WithField("extra", nodesToJoin).Info("Expand.")
 
-	for _, node := range extra {
-		c.Logger().WithField("node", node).Info("Join.")
-		err = node.Join(ctx, JoinCmd{
-			PeerAddr: joinAddr,
-			Token:    status.Cluster.Token.Token,
-			Role:     p.Role,
-			StateDir: p.StateDir,
-		})
+	for _, node := range nodesToJoin {
+		err = c.joinNode(peer, node, token, p)
 		if err != nil {
 			return trace.Wrap(err, "error joining cluster on node %s: %v", node.String(), err)
 		}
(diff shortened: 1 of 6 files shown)


infra/gravity/cluster_resize.go
```diff
@@ -68,22 +68,24 @@ func (c *TestContext) ShrinkLeave(nodesToKeep, nodesToRemove []Gravity) error {
...
 
-	err := master.Remove(ctx, remove.Node().PrivateAddr(), Graceful(!remove.Offline()))
-	if err != nil {
-		return trace.Wrap(err)
+	master := nodesToKeep[0]
+
+	errs := make(chan error, len(nodesToRemove))
+	for _, node := range nodesToRemove {
+		go func(n Gravity) {
+			err := master.Remove(ctx, n.Node().PrivateAddr(), Graceful(!n.Offline()))
```
r0mant, line 85: Seems like it can now launch multiple "shrink" operations simultaneously? Gravity actually doesn't support that, so if there's an ongoing shrink operation, subsequent "gravity remove" commands will exit with error until the current one completes. I think that's the reason this only accepted a single node. It might be fine to keep this updated interface in robotest for future when (or if) we actually start allowing simultaneous shrinks (or tweak remove command so it waits till the current operation completes).

infra/gravity/cluster_resize.go
```diff
@@ -68,22 +68,24 @@ func (c *TestContext) ShrinkLeave(nodesToKeep, nodesToRemove []Gravity) error {
 	return trace.Wrap(utils.CollectErrors(ctx, errs))
 }
 
-// RemoveNode simulates sudden nodes loss within an existing cluster followed by node eviction
-func (c *TestContext) RemoveNode(nodesToKeep []Gravity, remove Gravity) error {
+// RemoveNodes evicts several nodes from the cluster
```
r0mant, line 71: The previous comment mentioned a "sudden node loss" but it seems like it's just doing regular "gravity remove"?

suite/sanity/shrink.go
```diff
@@ -0,0 +1,47 @@
+package sanity
```
r0mant, line 1: Missing license header.

suite/sanity/loss_recover.go
```diff
@@ -104,9 +104,11 @@ func lossAndRecovery(p interface{}) (gravity.TestFunc, error) {
 			g.Logger().WithFields(logrus.Fields{"roles": roles, "nodes": nodes}).
 				Info("roles after expand")
 
-			g.OK("remove old node", g.RemoveNode(nodes, removed))
+			g.OK("remove old node", g.RemoveNodes(nodes, []gravity.Gravity{removed}))
```
r0mant, line 107: Nit: Not really sure why it says "old" node, I'd just say "a" node.

suite/sanity/shrink.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/gravitational/robotest/infra/gravity"
+
+	"github.com/sirupsen/logrus"
+)
+
+func shrink(p interface{}) (gravity.TestFunc, error) {
```
r0mant, line 9: So I'm curious about the UX of this. We already have a "resize" scenario which AFAIU is (was) supposed to mean both "expand" and "shrink". So if you resize 1->3, this means "expand", while e.g. 3->2 is "shrink".

I wonder if you have you considered augmenting the resize scenario to support node removal (as it should I think) instead of adding a separate "shrink" case?

infra/gravity/cluster_resize.go
wadells: The previous version of this wasn't doing a sudden node loss either.  The way that `loss_recover` calls this function conditionally would power off the node before hand.  Either way, the previous comment was misleading, which is why I updated the comment.

infra/gravity/cluster_resize.go
wadells: Thanks! I didn't realize that.  I'll serialize this loop and update the timeouts accordingly.

suite/sanity/shrink.go
wadells: Including a license header is not currently a standard in the robotest code base.  I've filed https://github.com/gravitational/robotest/issues/231 as a followup work item to add it to this (and all other) file(s).

suite/sanity/shrink.go
wadells: "resize" claimed to shrink, but it does not (currently) do so in practice. `recover` implements shrink and expand with several configurable parameters, and is in a state of disrepair. I considered repairing either or both of these as the way forward with shrink.  It would not be difficult to transition this work over to either.

To address "Why did you chose a new test case?"

As I see them, Robotest's largest pain points are:

1. determinism (developers don't trust results)
2. speed (it takes a long time for PR builds to compete)
3. development/contribution complexity (no one has the priority/wants to touch it because it is difficult and error prone to do so).

Implementing configurable test cases with structural branching tends to hurt us in 1 & 3.  That being said, we'll always have a balance within robotest (e.g. writing a separate test case for a different distro or number of nodes in the cluster feels foolish).  Currently, I see our dividing line as:

Items that affect the initial gravity cluster configuration (the neutral state at the start of the scenario we want to exercise) should be parameterized.  Items that determine what exactly is exercised during the scenario should be coded into the test code.  Following this out, I argue both `resize` and `recover` should be refactored into smaller independent test cases instead of expanded to cover more scenarios. `resize` -> `expand` and `shink`. `recover` into several independent `recoverX` (e.g. `recoverMaster`, `recoverKnode`) for the different scenarios that are of interest.

For some 3rd party thought on the situation I recommend some of Google's TOTT:

https://testing.googleblog.com/2018/06/testing-on-toilet-keep-tests-focused.html
https://testing.googleblog.com/2014/04/testing-on-toilet-test-behaviors-not.html

P.S. Getting off on a tangent, much of the provisioning that is currently implemented inline in tests should be factored out into fixtures -- which will provide further distinction between "something in the cloud went wrong" and "the behavior we were trying to test failed".  This data could then feed into metrics about utility: What is the ratio of ERR to FAIL?  Given our $$ estimates per scenario, where are we best served by investment? Furthermore, with provisioning factored out of test cases, the parameterization would (almost?) entirely be for the provisioner, and test code would simply receive pointers to the nodes the need, ready to go.  These ideas are non-trivial refactors of robotests internal harness... so I'm not going to further delay shrink or other near-term objectives while I explore them.

r0mant: Okay, sounds good. I was also thinking that we're currently lacking an ability to easily combine all our individual test-cases into more complicated scenarios, our current scenarios are pretty rigid. I.e. as implemented, "shrink" scenario also tests join even though we also have a separate scenario for it as well.

I think that ideally all operations we support should be sort of "building blocks" that robotest understands and then we can construct test scenarios easily in declarative way.

Regarding "resize" vs "expand"/"shrink" - it's probably just a question of semantics, but I think that it'd be nice to try and be consistent, i.e. if we have "shrink" then we probably also should have "expand" (esp. since "resize" really is basically expand). We don't need to do it in this PR though.

infra/gravity/cluster_resize.go
wadells: Reworked in 1ed898d.

suite/sanity/loss_recover.go
wadells: Cleaned up in 1ed898d.

suite/sanity/shrink.go
wadells: > I.e. as implemented, "shrink" scenario also tests join even though we also have a separate scenario for it as well.

The join-before-shrink is 100% because I'm trying to regression test gravitational/gravity#1445. My initial shrink implementation did not have a join after the install (which failed to catch 1445 because it is dependent on the extra `knode` join).

I considered implementing this as two test cases, one as a pure shrink and a 2nd as a regression test specific to 1445.  This approach is "cleaner" test wise.  However, we're already running pretty hot in terms of robotest cloud spend::bugs caught, so I was reluctant to have two test cases when one got us most of the value.  This argument is a lot weaker since 1445 doesn't consistently repro with the current test case.  I'm still trying to figure out why -- but I wanted to get this up for review.

> I think that ideally all operations we support should be sort of "building blocks" that robotest understands and then we can construct test scenarios easily in declarative way.

As long as we keep the test cases in the robotest repo, our building blocks are `infra/*` (should probably be renamed lib at some point) and our test cases are `suite/sanity/*`.  As is, `suite/sanity/*` are not declarative, but they're getting closer.

If, however, we move the test cases into Gravity, having robotest be a library of building blocks would make a ton of sense. There is considerable merit having the test case definition live in Gravity, most notably: it becomes easier for gravity developers to contribute new test cases., and test cases merge/revert/whatever with other Gravity code. If we pursue this, we'll almost certainly want to library-ize robotest and find a way to have Gravity's test definitions in go.  We currently do bash -> json(-ish) config -> go -> terraform .  We can cut out the bash & json portion if we turn robotest into a library.  I don't think adding further complexity our bash & json-ish DSL is a sustainable future. 

Following this to its conclusion: Does it make sense to have robotest code as a separate repo?  Probably, but it may start to be treated more like planet.

PR #233, opened 2020-05-28
Check for "active" status

### LINKED ISSUE #225: Assert more about gravity status than 'not degraded' after install
### Summary
After robotest runs an install, it checks `gravity status` is not 'degraded'.

This allows bugs like https://github.com/gravitational/gravity/issues/1523 to escape the install test (though they may be caught by other tests which run post-install operations)

### Expected Behavior
I expect our robotest install tests to catch status issues like https://github.com/gravitational/gravity/issues/1523 

### Logs
An example failure (during an in development shrink test) can be seen [here](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-13T00:08:51.568000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22776556d1-2a52-4168-a29d-17ff8aa5511d%22%0Alabels.__suite__%3D%22248439f5-9cfc-459d-b5eb-143b5d7c44ba%22&dateRangeEnd=2020-05-13T00:08:48.292Z&interval=PT1H&scrollTimestamp=2020-05-12T20:53:49.127493095Z&dateRangeUnbound=backwardInTime).

### Additional Context
This bug came from the discussion here: https://github.com/gravitational/gravity/pull/1537#issuecomment-627704884

wadells:
## Description
This change introduces more rigorous status checking in all suite tests cases.  Specifically, we go from `state != degraded` (30m timeout) to `state = active && system_status = Running` (5m timeout).

## Type of change
* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* Internal change (not necessarily a bug fix or a new feature)

I mention breaking because the 30m -> 5m timeout may break some of the tests cases we don't run presently.  It should be safe for all currently run test cases, because they wait for operations to complete before querying status.

## Linked tickets and other PRs

Fixes #225.
Heavily rewrites the code introduced in #136.

## TODOs
- [x] Self-review the change
- [x] Write documentation
- [x] Address review feedback
- [x] Write automated tests

Manual testing
- [x] install testing
- [x] upgrade testing
- [x] shrink testing
- [x] resize/expand testing

## Testing done
Added unit tests for status validations. 

Logs from end-to-end runs:
* [upgrade](https://console.cloud.google.com/logs/viewer?advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%227716b2b9-495f-4994-b9df-4ed8b1eca546%22%0Alabels.__suite__%3D%2286f73a17-e7ed-4c5a-8e18-752345982871%22%0Aseverity%3E%3DINFO&authuser=0&expandAll=false&project=kubeadm-167321)
* [install](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-06-04T17:29:38.218000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%2231641fab-7122-405a-bc8f-661bc6ce3dcf%22%0Alabels.__suite__%3D%2218283bab-ebfc-4f94-851e-4e8b52e79a40%22%0Aseverity%3E%3DINFO&dateRangeStart=2020-06-04T16:29:38.731Z&dateRangeEnd=2020-06-04T17:29:38.731Z&interval=PT1H&scrollTimestamp=2020-06-04T17:29:08.893390113Z) (updated after round 2 of changes)
* [shrink](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-28T20:20:45.516000000Z&customFacets=&limitCustomFacetWidth=true&interval=CUSTOM&scrollTimestamp=2020-05-27T20:56:28.810552801Z&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22b32c7b8c-931b-4fb7-8f35-86e319db8ff2%22%0Alabels.__suite__%3D%225199396c-9a77-4d8c-8a95-4cc93b923652%22%0Aseverity%3E%3DINFO&dateRangeEnd=2020-05-28T20:20:43.140Z&dateRangeUnbound=backwardInTime) (updated after review feedback)
* [resize/expand](https://console.cloud.google.com/logs/viewer?advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%228eba6e3a-8072-4cd4-a519-84dddba1ac8c%22%0Alabels.__suite__%3D%2261652cea-69b5-4c70-b99c-d5396fe883a1%22%0Aseverity%3E%3DINFO&authuser=0&expandAll=false&project=kubeadm-167321) (updated after review feedback)

### DIFF
--- a/infra/gravity/cluster_install.go
+++ b/infra/gravity/cluster_install.go
@@ -232,7 +232,7 @@ func (c *TestContext) upgrade(master Gravity, numNodes int) error {
 
 // ExecScript will run and execute a script on all nodes
 func (c *TestContext) ExecScript(nodes []Gravity, scriptUrl string, args []string) error {
-	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
+	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.ExecScript)
 	defer cancel()
 
 	errs := make(chan error, len(nodes))
--- a/infra/gravity/cluster_install.go
+++ b/infra/gravity/cluster_install.go
@@ -205,7 +205,7 @@ func (c *TestContext) uploadInstaller(master Gravity, nodes []Gravity, installer
 		return trace.Wrap(err)
 	}
 
-	err = c.Status(nodes)
+	err = c.WaitForActiveStatus(nodes)
 	if err != nil {
 		return trace.Wrap(err)
 	}
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -17,8 +17,8 @@ type statusValidator func(s GravityStatus) error
 
 // checkActive returns an error if the cluster is degraded or state != active.
 func checkActive(s GravityStatus) error {
-	if s.Cluster.Status != constants.ClusterStateActive {
-		return trace.CompareFailed("expected state %q, found %q", constants.ClusterStateActive, s.Cluster.Status)
+	if s.Cluster.State != constants.ClusterStateActive {
+		return trace.CompareFailed("expected state %q, found %q", constants.ClusterStateActive, s.Cluster.State)
 	}
 	return nil
 }
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -53,9 +53,8 @@ func (c *TestContext) CheckTimeSync(nodes []Gravity) error {
 		})
 	}
 
-	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
+	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.TimeSync)
 	defer cancel()
-
 	err := sshutils.CheckTimeSync(ctx, timeNodes)
 	return trace.Wrap(err)
 }
@@ -122,7 +121,7 @@ type ClusterNodesByRole struct {
 
 // NodesByRole will conveniently organize nodes according to their roles in cluster
 func (c *TestContext) NodesByRole(nodes []Gravity) (roles *ClusterNodesByRole, err error) {
-	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
+	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.ResolveInPlanet)
 	defer cancel()
 
 	roles = &ClusterNodesByRole{}
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -15,8 +15,24 @@ import (
 // statusValidator returns nil if the Gravity Status is the expected status or an error otherwise.
 type statusValidator func(s GravityStatus) error
 
+// checkNotDegraded returns an error if the cluster status is Degraded.
+//
+// This function is a reimplementation of the logic in https://github.com/gravitational/gravity/blob/7.0.0/lib/status/status.go#L180-L185
+func checkNotDegraded(s GravityStatus) error {
+	if s.Cluster.State == constants.ClusterStateDegraded {
+		return trace.CompareFailed("cluster state %q", s.Cluster.State)
+	}
+	if s.Cluster.SystemStatus != constants.SystemStatus_Running {
+		return trace.CompareFailed("expected system_status %v, found %v", constants.SystemStatus_Running, s.Cluster.SystemStatus)
+	}
+	return nil
+}
+
 // checkActive returns an error if the cluster is degraded or state != active.
 func checkActive(s GravityStatus) error {
+	if err := checkNotDegraded(s); err != nil {
+		return trace.Wrap(err)
+	}
 	if s.Cluster.State != constants.ClusterStateActive {
 		return trace.CompareFailed("expected state %q, found %q", constants.ClusterStateActive, s.Cluster.State)
 	}
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -2,45 +2,78 @@ package gravity
 
 import (
 	"context"
-	"time"
 
+	"github.com/cenkalti/backoff"
+
+	"github.com/gravitational/robotest/lib/defaults"
 	sshutils "github.com/gravitational/robotest/lib/ssh"
 	"github.com/gravitational/robotest/lib/utils"
 	"github.com/gravitational/robotest/lib/wait"
-
 	"github.com/gravitational/trace"
 )
 
-// Status walks around all nodes and checks whether they all feel OK
-func (c *TestContext) Status(nodes []Gravity) error {
-	c.Logger().WithField("nodes", Nodes(nodes)).Info("Check status on nodes.")
-	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
-	defer cancel()
+// WaitForActiveStatus blocks until all nodes report state = Active and notDegraded or an internal timeout expires.
+func (c *TestContext) WaitForActiveStatus(nodes []Gravity) error {
+	c.Logger().WithField("nodes", Nodes(nodes)).Info("Waiting for active status.")
+	return c.WaitForStatus(nodes, defaults.GravityStateActive)
+}
 
-	retry := wait.Retryer{
-		Attempts: 100,
-		Delay:    time.Second * 20,
+// WaitForStatus blocks until all nodes satisfy the expected statusValidator or an internal timeout expires.
+func (c *TestContext) WaitForStatus(nodes []Gravity, expected string) error {
+	b := backoff.NewExponentialBackOff()
+	b.MaxElapsedTime = c.timeouts.ClusterStatus
+
+	expectStatus := func() (err error) {
+		statuses, err := c.Status(nodes)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		for _, status := range statuses {
+			if status.Cluster.Status != expected {
+				c.Logger().WithError(err).WithField("status", status).Warn("Unexpected Status.")
+				return trace.CompareFailed("expected status %q, found %q", expected, status.Cluster.Status)
+			}
+		}
+		return nil
 	}
 
-	err := retry.Do(ctx, func() error {
-		errs := make(chan error, len(nodes))
-
-		for _, node := range nodes {
-			go func(n Gravity) {
-				_, err := n.Status(ctx)
-				errs <- err
-			}(node)
-		}
-
-		err := utils.CollectErrors(ctx, errs)
-		if err == nil {
-			return nil
-		}
-		c.Logger().Warnf("Status not available on some nodes, will retry: %v.", err)
-		return wait.Continue("status not ready on some nodes")
-	})
+	err := wait.RetryWithInterval(c.ctx, b, expectStatus, c.Logger())
 
 	return trace.Wrap(err)
+
+}
+
+// Status queries `gravity status` on each node in nodes.
+func (c *TestContext) Status(nodes []Gravity) ([]GravityStatus, error) {
+	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.NodeStatus)
+	defer cancel()
+
+	errC := make(chan error, len(nodes))
+	valueC := make(chan interface{}, len(nodes))
+
+	for _, node := range nodes {
+		go func(n Gravity) {
+			status, err := n.Status(ctx)
+			errC <- err
+			if status != nil {
+				valueC <- *status
+			}
+		}(node)
+	}
+
+	values, err := utils.Collect(ctx, cancel, errC, valueC)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	var statuses []GravityStatus
+	for _, v := range values {
+		status, ok := v.(GravityStatus)
+		if !ok {
+			return nil, trace.BadParameter("expected %T, got %T", status, v)
+		}
+		statuses = append(statuses, status)
+	}
+	return statuses, nil
 }
 
 // CheckTime walks around all nodes and checks whether their time is within acceptable limits
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -5,21 +5,32 @@ import (
 
 	"github.com/cenkalti/backoff"
 
-	"github.com/gravitational/robotest/lib/defaults"
+	"github.com/gravitational/robotest/lib/constants"
 	sshutils "github.com/gravitational/robotest/lib/ssh"
 	"github.com/gravitational/robotest/lib/utils"
 	"github.com/gravitational/robotest/lib/wait"
 	"github.com/gravitational/trace"
 )
 
+// statusValidator returns nil if the Gravity Status is the expected status or an error otherwise.
+type statusValidator func(s GravityStatus) error
+
+// checkActive returns an error if the cluster is degraded or state != active.
+func checkActive(s GravityStatus) error {
+	if s.Cluster.Status != constants.ClusterStateActive {
+		return trace.CompareFailed("expected state %q, found %q", constants.ClusterStateActive, s.Cluster.Status)
+	}
+	return nil
+}
+
 // WaitForActiveStatus blocks until all nodes report state = Active and notDegraded or an internal timeout expires.
 func (c *TestContext) WaitForActiveStatus(nodes []Gravity) error {
 	c.Logger().WithField("nodes", Nodes(nodes)).Info("Waiting for active status.")
-	return c.WaitForStatus(nodes, defaults.GravityStateActive)
+	return c.WaitForStatus(nodes, checkActive)
 }
 
 // WaitForStatus blocks until all nodes satisfy the expected statusValidator or an internal timeout expires.
-func (c *TestContext) WaitForStatus(nodes []Gravity, expected string) error {
+func (c *TestContext) WaitForStatus(nodes []Gravity, expected statusValidator) error {
 	b := backoff.NewExponentialBackOff()
 	b.MaxElapsedTime = c.timeouts.ClusterStatus
 
@@ -29,9 +40,10 @@ func (c *TestContext) WaitForStatus(nodes []Gravity, expected string) error {
 			return trace.Wrap(err)
 		}
 		for _, status := range statuses {
-			if status.Cluster.Status != expected {
+			err = expected(status)
+			if err != nil {
 				c.Logger().WithError(err).WithField("status", status).Warn("Unexpected Status.")
-				return trace.CompareFailed("expected status %q, found %q", expected, status.Cluster.Status)
+				return trace.Wrap(err)
 			}
 		}
 		return nil
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -130,6 +130,8 @@ func (c *TestContext) NodesByRole(nodes []Gravity) (roles *ClusterNodesByRole, e
 		return nil, trace.Wrap(err)
 	}
 
+	ctx, cancel = context.WithTimeout(c.ctx, c.timeouts.GetPods)
+	defer cancel()
 	// Run query on the apiserver
 	pods, err := KubectlGetPods(ctx, roles.ApiMaster, kubeSystemNS, appGravityLabel)
 	if err != nil {
--- a/infra/gravity/cluster_status.go
+++ b/infra/gravity/cluster_status.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/cenkalti/backoff"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/gravitational/robotest/lib/constants"
 	sshutils "github.com/gravitational/robotest/lib/ssh"
@@ -71,34 +72,32 @@ func (c *TestContext) WaitForStatus(nodes []Gravity, expected statusValidator) e
 
 }
 
-// Status queries `gravity status` on each node in nodes.
-func (c *TestContext) Status(nodes []Gravity) ([]GravityStatus, error) {
+// Status queries `gravity status` once from each node in nodes.
+func (c *TestContext) Status(nodes []Gravity) (statuses []GravityStatus, err error) {
 	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.NodeStatus)
 	defer cancel()
 
-	errC := make(chan error, len(nodes))
-	valueC := make(chan interface{}, len(nodes))
-
+	valueC := make(chan GravityStatus, len(nodes))
+	g, ctx := errgroup.WithContext(ctx)
 	for _, node := range nodes {
-		go func(n Gravity) {
-			status, err := n.Status(ctx)
-			errC <- err
+		node := node
+		g.Go(func() error {
+			status, err := node.Status(ctx)
+			if err != nil {
+				return trace.Wrap(err)
+			}
 			if status != nil {
 				valueC <- *status
 			}
-		}(node)
+			return nil
+		})
 	}
-
-	values, err := utils.Collect(ctx, cancel, errC, valueC)
+	err = g.Wait()
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	var statuses []GravityStatus
-	for _, v := range values {
-		status, ok := v.(GravityStatus)
-		if !ok {
-			return nil, trace.BadParameter("expected %T, got %T", status, v)
-		}
+	close(valueC)
+	for status := range valueC {
 		statuses = append(statuses, status)
 	}
 	return statuses, nil
--- a/infra/gravity/gravity_test.go
+++ b/infra/gravity/gravity_test.go
@@ -3,6 +3,7 @@ package gravity
 import (
 	"bufio"
 	"bytes"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -14,10 +15,11 @@ func TestGravityOutput(t *testing.T) {
 `)
 	expectedStatus := &GravityStatus{
 		Cluster: ClusterStatus{
-			Cluster:     "testcluster",
-			Application: Application{Name: "telekube"},
-			State:       "active",
-			Token:       Token{Token: "fac3b88014367fe4e98a8664755e2be4"},
+			Cluster:      "testcluster",
+			Application:  Application{Name: "telekube"},
+			State:        "active",
+			SystemStatus: 1,
+			Token:        Token{Token: "fac3b88014367fe4e98a8664755e2be4"},
 			Nodes: []NodeStatus{
 				NodeStatus{Addr: "10.40.2.4"},
 				NodeStatus{Addr: "10.40.2.5"},
@@ -38,10 +40,11 @@ func TestGravityOutput(t *testing.T) {
 func TestHealthyStatusValidation(t *testing.T) {
 	healthyStatus := GravityStatus{
 		Cluster: ClusterStatus{
-			Cluster:     "robotest",
-			Application: Application{Name: "telekube"},
-			State:       "active",
-			Token:       Token{Token: "ROBOTEST"},
+			Cluster:      "robotest",
+			Application:  Application{Name: "telekube"},
+			State:        "active",
+			SystemStatus: 1,
+			Token:        Token{Token: "ROBOTEST"},
 			Nodes: []NodeStatus{
 				NodeStatus{Addr: "10.1.2.3"},
 				NodeStatus{Addr: "10.1.2.4"},
@@ -63,7 +66,7 @@ func Test1523StatusValidation(t *testing.T) {
 		Cluster: ClusterStatus{
 			Cluster:     "robotest",
 			Application: Application{Name: "telekube"},
-			State:       "degraded",
+			State:       "expanding",
 			Token:       Token{Token: "ROBOTEST"},
 			Nodes: []NodeStatus{
 				NodeStatus{Addr: "10.1.2.3"},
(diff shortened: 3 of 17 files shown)


infra/gravity/cluster_install.go
```diff
@@ -205,7 +205,7 @@ func (c *TestContext) uploadInstaller(master Gravity, nodes []Gravity, installer
 		return trace.Wrap(err)
 	}
 
-	err = c.Status(nodes)
+	_, err = c.Status(nodes)
```
wadells, line 208: This previously looped, waiting for `!= degraded`.  Now it will check only once.  We can't use to `c.WaitForActiveStatus(nodes)` because I don't believe that `gravity status` returns active before the install.  Perhaps we need another wait function?

infra/gravity/cluster_install.go
```diff
@@ -232,7 +232,7 @@ func (c *TestContext) upgrade(master Gravity, numNodes int) error {
 
 // ExecScript will run and execute a script on all nodes
 func (c *TestContext) ExecScript(nodes []Gravity, scriptUrl string, args []string) error {
-	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
+	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.ExecScript)
```
wadells, line 235: As mentioned below, this timeout should be specified by the end user who specifies the exec script.  Needs a followup ticket.

lib/constants/constants.go
```diff
@@ -54,3 +54,16 @@ const (
...
+const (
+	// SiteState come from https://github.com/gravitational/gravity/blob/7.0.0/lib/ops/constants.go#L64-L93
+	SiteStateActive   = "active"
+	SiteStateDegraded = "degraded"
+
+	// SystemStatus come from https://github.com/gravitational/satellite/blob/7.1.0/agent/proto/agentpb/agent.pb.go#L28-L32
+	SystemStatus_Unknown  = 0
+	SystemStatus_Running  = 1
+	SystemStatus_Degraded = 2
+)
```
wadells, line 69: Should these be in lib?  Seems like `infra/gravity/constants` or similar may be a better home.  I need to look into what the dividing line between `lib` and `infra` is intended to be.

suite/sanity/autoscale.go
```diff
@@ -15,19 +15,20 @@ func autoscale(p interface{}) (gravity.TestFunc, error) {
...
+		g.OK("wait for active status on masters", g.WaitForActiveStatus(cluster.Nodes))
+		g.OK("wait for active status on asg workers", g.WaitForActiveStatus(workers))
 
 		// Scale Down
 		workers, err = g.AutoScale(1)
 		g.OK("asg-down", err)
-		g.OK("status-masters", g.Status(cluster.Nodes))
-		g.OK("status-workers", g.Status(workers))
+		g.OK("wait for active status on masters", g.WaitForActiveStatus(cluster.Nodes))
+		_, err = g.Status(workers)
```
wadells, line 31: This no longer waits for `state != degraded`.  We don't run this test presently, but if we did this could be a breaking change.

infra/gravity/cluster_install.go
wadells: This is responsible for the upgrade failure seen [here](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-28T23:07:45.203000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22dbc09cee-cb33-4304-993e-7db324cec2a0%22%0Alabels.__suite__%3D%22193da8e2-b2cc-4439-8855-b7f80c6ae3a2%22&dateRangeStart=2020-05-28T22:07:57.117Z&dateRangeEnd=2020-05-28T23:07:57.117Z&interval=PT1H&scrollTimestamp=2020-05-28T22:41:17.837368627Z).  Fixed in e4c7185

infra/gravity/cluster_install.go
wadells: Followup PR: https://github.com/gravitational/robotest/pull/235

suite/sanity/autoscale.go
wadells: I'm going to ignore this until we have occasion to bring autoscale back.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+// a negation / reimplementation of https://github.com/gravitational/gravity/blob/7.0.0/lib/status/status.go#L180-L185
+// I'd prefer isDegraded and negation later, but I didn't find an elegant way to
+// have consumer friendly error messages when the information channel is squashed to != nil
+// -- 2020-05 walt
+func isNotDegraded(s GravityStatus) error {
+	if s.Cluster.State == constants.SiteStateDegraded {
+		return trace.Errorf("cluster state %q", s.Cluster.State)
+	}
+	if s.Cluster.SystemStatus != constants.SystemStatus_Running {
+		return trace.Errorf("expected system_status %v, found %v", constants.SystemStatus_Running, s.Cluster.SystemStatus)
```
a-palchikov, line 27: `trace.CompareFailed`

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
-	c.Logger().WithField("nodes", Nodes(nodes)).Info("Check status on nodes.")
-	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
-	defer cancel()
+// returns nil if the Gravity Status is the expected status or an error otherwise
+type statusValidator func(s GravityStatus) error
 
-	retry := wait.Retryer{
-		Attempts: 100,
-		Delay:    time.Second * 20,
+// a negation / reimplementation of https://github.com/gravitational/gravity/blob/7.0.0/lib/status/status.go#L180-L185
```
a-palchikov, line 18: I don't follow your comment - can you clarify what's missing or expected on the gravity side?

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
-				errs <- err
-			}(node)
-		}
+// not degraded and state == active
+func isActive(s GravityStatus) error {
+	if err := isNotDegraded(s); err != nil {
+		return trace.Wrap(err)
+	}
+	if s.Cluster.State != constants.SiteStateActive {
+		return trace.Errorf("expected state %q, found %q", constants.SiteStateActive, s.Cluster.State)
```
a-palchikov, line 39: `trace.CompareFailed`

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+	}
+	return nil
+}
+
+func (c *TestContext) WaitForActiveStatus(nodes []Gravity) error {
```
a-palchikov, line 44: `// WaitForActiveStatus blocks until the nodes specified with nodes all report healthy or internal context expires`?
What's the timeout condition - can you clarify?

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+	c.Logger().WithField("nodes", Nodes(nodes)).Info("Waiting for active status.")
+	return c.WaitForStatus(nodes, isActive)
+}
+
+// Waits for all nodes to be in the same status, returns an error if nodes do not converge
```
a-palchikov, line 49: The method comment should start with the method name, i.e.:

`// WaitForStatus ... `

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+
+	err := wait.RetryWithInterval(c.ctx, b, expectStatus, c.Logger())
 
 	return trace.Wrap(err)
+
+}
+
+// Status queries the status on all nodes.
+// Returns an partial status info and an error if any individual node status queries fail.
+func (c *TestContext) Status(nodes []Gravity) ([]*GravityStatus, error) {
```
a-palchikov, line 82: Do you need `GravityStatus` as pointers? It's more hassle if they can never be nil.

infra/gravity/gravity_test.go
```diff
@@ -35,3 +36,53 @@ func TestGravityOutput(t *testing.T) {
...
+				NodeStatus{Addr: "10.1.2.5"},
+			},
+		},
+	}
+	err := isActive(healthyStatus)
+	assert.NoError(t, err)
+}
+
+func TestIsActiveStatusValidation(t *testing.T) {
+	unhealthyStatus := GravityStatus{
```
a-palchikov, line 60: Why is this unhealthy?

infra/gravity/gravity_test.go
```diff
@@ -35,3 +36,53 @@ func TestGravityOutput(t *testing.T) {
...
+	assert.Error(t, err)
+}
+
+// Status may be internally inconsistent. Do not consider
+// status like these active & ready to proceed. See:
+//   https://github.com/gravitational/gravity/issues/1521
+func Test1521StatusValidation(t *testing.T) {
+	f, err := os.Open("testdata/status-degraded-1521.json")
+	assert.NoError(t, err)
+
```
a-palchikov, line 81: `defer f.Close()`

infra/gravity/testdata/status-degraded-1521.json
```diff
@@ -0,0 +1,129 @@
+{
+  "cluster": {
+    "application": {
+      "repository": "gravitational.io",
+      "name": "telekube",
+      "version": "6.1.27"
+    },
+    "state": "active",
```
a-palchikov, line 8: The [1521](https://github.com/gravitational/gravity/issues/1521) is about a different situation. The cluster state should never be active with failed probes (as in this case below). The issue is about the state when the cluster is healthy (i.e. no failed probes) but the cluster controller (the active `gravity-site` Pod) did not reset the cluster state to `active` since its own state checker loop has yet to determine the cluster state.

suite/sanity/resize.go
```diff
@@ -41,11 +41,11 @@ func resize(p interface{}) (gravity.TestFunc, error) {
...
 		g.OK(fmt.Sprintf("install on %d node", param.NodeCount),
 			g.OfflineInstall(cluster.Nodes[0:param.NodeCount], param.InstallParam))
-		g.OK("status", g.Status(cluster.Nodes[0:param.NodeCount]))
+		g.OK("wait for active status", g.WaitForActiveStatus(cluster.Nodes[0:param.NodeCount]))
 		g.OK("time sync", g.CheckTimeSync(cluster.Nodes))
 		g.OK(fmt.Sprintf("expand to %d nodes", param.ToNodes),
 			g.Expand(cluster.Nodes[0:param.NodeCount], cluster.Nodes[param.NodeCount:param.ToNodes],
 				param.InstallParam))
-		g.OK("status", g.Status(cluster.Nodes[0:param.ToNodes]))
+		g.OK("wait for active status", g.WaitForActiveStatus(cluster.Nodes[0:param.ToNodes]))
```
a-palchikov, line 49: nit: `0` is redundant here

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
-		err := utils.CollectErrors(ctx, errs)
-		if err == nil {
-			return nil
+	expectStatus := func() (err error) {
+		statuses, err := c.Status(nodes)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		for _, sp := range statuses {
+			if sp == nil { // should never trigger, as the above err != nil check catches this path on well shaped data
```
a-palchikov, line 60: This should never be part of the slice then. I also commented on the use of pointers vs values - we don't need pointers here, and if you used values, there would be no need for `nil` test.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
-
 	"github.com/gravitational/trace"
 )
 
-// Status walks around all nodes and checks whether they all feel OK
-func (c *TestContext) Status(nodes []Gravity) error {
-	c.Logger().WithField("nodes", Nodes(nodes)).Info("Check status on nodes.")
-	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.Status)
-	defer cancel()
+// returns nil if the Gravity Status is the expected status or an error otherwise
```
r0mant, line 15: Nit: We follow Go Code Review Comments style guide for our code, according to which Godoc strings should start with the name of the thing they describe. I.e. `statusValidator declares a function type that returns nil if ....`

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

Same for the Godoc strings below.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+type statusValidator func(s GravityStatus) error
 
-	retry := wait.Retryer{
-		Attempts: 100,
-		Delay:    time.Second * 20,
+// a negation / reimplementation of https://github.com/gravitational/gravity/blob/7.0.0/lib/status/status.go#L180-L185
+// I'd prefer isDegraded and negation later, but I didn't find an elegant way to
```
r0mant, line 19: Nit: To be honest I'm not sure about the value of comments like this, it reads more like a personal journal :) I would either remove this, or replace with a specific `TODO(walt)` item (which is the format we usually use for things like that).

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+// have consumer friendly error messages when the information channel is squashed to != nil
+// -- 2020-05 walt
+func isNotDegraded(s GravityStatus) error {
```
r0mant, line 22: Nit: Speaking about semantics, I think that functions named like "isXXX" are rather expected to return a boolean value. Otherwise it might be a bit confusing i.e. does this return an error if it's degraded or not?

I would consider renaming this to "checkNotDegraded" or "assertNotDegraded".

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+	if s.Cluster.State == constants.SiteStateDegraded {
+		return trace.Errorf("cluster state %q", s.Cluster.State)
```
r0mant, line 24: Similar to Dima's comment below - we try to avoid using `trace.Errorf` directly, instead use specific errors from trace package. Here `trace.CompareFailed` seems to fit well.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+}
 
-		for _, node := range nodes {
-			go func(n Gravity) {
-				_, err := n.Status(ctx)
-				errs <- err
-			}(node)
-		}
+// not degraded and state == active
+func isActive(s GravityStatus) error {
```
r0mant, line 34: Same comment as above - consider renaming to `[check|assert]Active`?

Also reminder about the Godoc string also from above.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+		return nil
+	}
+
+	err := wait.RetryWithInterval(c.ctx, b, expectStatus, c.Logger())
 
 	return trace.Wrap(err)
+
+}
+
+// Status queries the status on all nodes.
```
r0mant, line 80: Does this query `gravity status` from each node? Or some other status? Maybe the comment needs clarification.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+		}(node)
+	}
+
+	values, err := utils.Collect(ctx, cancel, errC, valueC)
+	// handling err is intentionally postponed, to accommodate some nodes returning status
+	// and some erroring. The below for loop *should* be safe even in the case of an err
+	// and the additional status data may be useful to the caller. -- 2020-05 walt
+	var statuses []*GravityStatus
+	for _, v := range values {
+		status := v.(*GravityStatus)
```
r0mant, line 103: Process type assert error just in case, to avoid potential panic?

```go
status, ok := v.(*GravityStatus)
if !ok {
  return nil, trace.BadParameter("expected GravityStatus, got %T", v)
}
```

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,108 @@ package gravity
...
+			errC <- err
+			valueC <- status
+		}(node)
+	}
+
+	values, err := utils.Collect(ctx, cancel, errC, valueC)
+	// handling err is intentionally postponed, to accommodate some nodes returning status
+	// and some erroring. The below for loop *should* be safe even in the case of an err
+	// and the additional status data may be useful to the caller. -- 2020-05 walt
+	var statuses []*GravityStatus
```
r0mant, line 101: Pointers are unnecessary here, let's just have a `[]GravityStatus`. This will simplify code above too as Dima pointed out.

infra/gravity/node_commands.go
```diff
@@ -142,8 +137,10 @@ type ClusterStatus struct {
 	Application Application `json:"application"`
 	// Cluster is the name of the cluster
 	Cluster string `json:"domain"`
-	// Status is the cluster status
-	Status string `json:"state"`
+	// State is the application state
```
r0mant, line 140: Nit: It's actually cluster state, not application.

lib/constants/constants.go
```diff
@@ -54,3 +54,16 @@ const (
 	// Ops specifies a special cloud provider - a telekube Ops Center
 	Ops = "ops"
 )
+
+// Gravity API constants redeclared here to avoid extra dependencies solely to
+// get these values
+const (
+	// SiteState come from https://github.com/gravitational/gravity/blob/7.0.0/lib/ops/constants.go#L64-L93
+	SiteStateActive   = "active"
```
r0mant, line 62: I know this is how they are defined in Gravity code, but we're trying to get away from using "Site" in our names and use "Cluster" instead. Also, same comment about Godoc format from above applies here as well.

```go
// ClusterStateActive defines the state a Gravity cluster is in when it's healthy.
ClusterStateActive = "active"
// ClusterStateDegraded defines the state a Gravity cluster is in when it's having issues.
ClusterStateDegraded = "degraded"
...
```

(feel free to tweak actual godocs)

infra/gravity/testdata/status-degraded-1521.json
wadells: I will update the test and filename.

> The cluster state should never be active with failed probes (as in this case below). 

Are you aware of a gravity issue tracking the inconsistency seen in this status?  Since this isn't 1521 but it sounds like it is a real problem, I'd like to link back to the correct issue.

infra/gravity/cluster_status.go
wadells: Thanks.  I'll get these comments cleaned up.  I've also filed https://github.com/gravitational/robotest/issues/236 to help preempt this.

infra/gravity/testdata/status-degraded-1521.json
a-palchikov: No, I think there's a bit of misunderstanding. `gravity status` should already work as expected w.r.t cluster state being degraded with failed probes. What [1521](https://github.com/gravitational/gravity/issues/1521) additionally describes is when the cluster state is degraded although the agent (planet) state is healthy. Since `gravity status` pulls the agent's status directly, it evaluates it in the form of node statuses. Hence, it planet reports healthy there will be no failed probes in the `gravity status` output.
The problem is then that `gravity status` does not control the cluster's state field - active `gravity-site` Pod does and it can still keep the cluster `degraded` (even when it is actually healthy - according to agent info) since it has its own loop that uses the same planet agent.
If `gravity status` pulled the info from planet  before `gravity-site` Pod did and the cluster was previously degraded, `gravity status` still reads degraded from cluster state and only when `gravity-site`'s loop pulls the state and sees that it's healthy will it reset the cluster state.
As long as we're waiting for `gravity-site` to reset the state, `gravity status` shows `degraded` for no obvious reason.
Hope that explains it.

wadells: Given we've already gone back & forth a couple times, I'd like to discuss this real time - perhaps Monday?

> gravity status should already work as expected w.r.t cluster state being degraded with failed probes. 

I pulled the status json in this file from robotest/gravity 6.1.27 the wild:  see [this query](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-29T22:34:05.422000000Z&customFacets=&limitCustomFacetWidth=true&interval=PT1H&scrollTimestamp=2020-05-27T21:12:12.519955220Z&dateRangeEnd=2020-05-29T22:34:03.251Z&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22ea6b230b-a447-4fc7-8128-6a2f3f1a9dce%22%0Alabels.__suite__%3D%22a9601994-7ead-43f0-ba76-b7cc497c3660%22%0AjsonPayload.cmd%3D%22sudo%20gravity%20status%20--output%3Djson%20--system-log-file%3Dgravity-system.log%22%0AAND%20jsonPayload.message:%22%5C%22state%5C%22:%5C%22active%5C%22,%22%0AAND%20jsonPayload.message:%22%5C%22system_status%5C%22:2,%22&dateRangeUnbound=backwardInTime) for an example.  Strip `uuid` and `suite` from the query and there are further examples.

Is this observed status inconsistency between `state`, `system_status` and `nodes` a bug in gravity?

If the inconsitency is a gravity bug: Does we already have an issue #? I'd like to link it in the code. I can create one if there isn't an issue.

If the status here is not a gravity bug: I don't understand what your original comment https://github.com/gravitational/robotest/pull/233#discussion_r432571825 is suggesting.  We should probably talk about it in a higher bandwidth medium.

infra/gravity/cluster_status.go
wadells: Fixed in 28aab700.

infra/gravity/testdata/status-degraded-1521.json
wadells: Renamed to https://github.com/gravitational/gravity/issues/1641 in 936a6c62.

infra/gravity/cluster_status.go
wadells: Addressed in 28aab700.

suite/sanity/resize.go
wadells: This was latent in the code already.  As such, I removed it in a separate commit from my other work: f364d2905a21789be0d70ccee346cdb2c34a6534.

infra/gravity/cluster_status.go
```diff
@@ -90,20 +83,23 @@ func (c *TestContext) Status(nodes []Gravity) ([]*GravityStatus, error) {
 		go func(n Gravity) {
 			status, err := n.Status(ctx)
 			errC <- err
-			valueC <- status
+			valueC <- *status
```
wadells, line 86: This needs to be done conditionally.  If err, this will result in a nil pointer dereference.

wadells: Fixed in d540cb5.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,106 @@ package gravity
...
+
+	err := wait.RetryWithInterval(c.ctx, b, expectStatus, c.Logger())
 
 	return trace.Wrap(err)
+
+}
+
+// Status queries `gravity status` on each node in nodes.
+func (c *TestContext) Status(nodes []Gravity) ([]GravityStatus, error) {
+	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.NodeStatus)
```
a-palchikov, line 76: nit: just to clarify - `NodeStatus` configures the status timeout for the whole status operation - not individual nodes?
In case it's per node you might need to offset the total amount by the size of the cluster.

infra/gravity/cluster_status.go
```diff
@@ -2,45 +2,106 @@ package gravity
...
+	}
+
+	values, err := utils.Collect(ctx, cancel, errC, valueC)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	var statuses []GravityStatus
+	for _, v := range values {
+		status, ok := v.(GravityStatus)
+		if !ok {
```
a-palchikov, line 99: I know this has been already been discussed and I could just offer my two cents on this by saying I'd be fine with type-cast panicking here since this does not escape the function (technically it does - into a goroutine by that happens in the context of the parent function so is ok) scope and in my book this is a programming error that should be fixed - not logged.
But I think this is a flawed abstraction in the first place - i.e. the use of boxing with `interface{}` only to extract bits into a specific type later. I know there's no way around it with these APIs, but the whole function can be also implemented inline with proper type checking, for example with the use of `errgroup` (but not necessarily):
```go
import "golang.org/x/sync/errgroup"
...
func (c *TestContext) Status(nodes []Gravity) (statuses []GravityStatus, err error) {
	ctx, cancel := context.WithTimeout(c.ctx, c.timeouts.NodeStatus)
	defer cancel()

	valueC := make(chan GravityStatus, len(nodes))
	g, ctx := errgroup.WithContext(ctx)
	for _, node := range nodes {
		node := node
		g.Go(func() error {
			status, err := node.Status(ctx)
			if err != nil {
				return trace.Wrap(err)
			}
			if status != nil {
				valueC <- *status
			}
			return nil
		})
	}
	err = g.Wait()
	if err != nil {
		return nil, trace.Wrap(err)
	}
	close(valueC)
	for status := range valueC {
		statuses = append(statuses, status)
	}
	return statuses, nil
}
```

infra/gravity/node_commands.go
```diff
@@ -142,8 +137,10 @@ type ClusterStatus struct {
 	Application Application `json:"application"`
 	// Cluster is the name of the cluster
 	Cluster string `json:"domain"`
-	// Status is the cluster status
-	Status string `json:"state"`
+	// State is the cluster state
+	State string `json:"state"`
+	// SystemStatus is the cluster status, see https://github.com/gravitational/satellite/blob/7.1.0/agent/proto/agentpb/agent.proto#L50-L54
```
a-palchikov, line 142: With the bits of 7.0.x being copy-pasted here and there - have you considered vendoring from the gravity repository directly?

wadells: > have you considered vendoring from the gravity repository directly?

TLDR; Yes. It is complex to do so and didn't seem high ROI.

I half implemented  (got a working build with dep on 7.0) this solution before breaking off and returning to the preexisting "copy key elements" strategy.

My concerns are:

1) Gravity core isn't developed with the library use case in mind.

I don't particularly want to push for this expectation -- I don't see it as an effective use of our time or limited review focus to start considering this now. @r0mant talked a bit about this back in January, when I first stumbled on https://github.com/golang/go/issues/32945:

> at least for gravity, I think it would problematic to migrate to v-prefixed versions
> well, maybe not problematic but will certainly requires code changes for little gain
> but then again, this is probably more important for libraries you vendor (which gravity is not)
> (so things like trace/roundtrip/satellite and other libs we use)

2) I'm concerned  formal dependencies will create a loop or obtuse case in our dependency graph.

Gravity doesn't presently have robotest as a go dependency or vice versa.  However there is an docker container dependency from gravity -> robotest.  And in https://github.com/gravitational/robotest/pull/229#discussion_r429019626 I explored the idea of having Gravity's tests depend explicitly on Robotest (via native golang API calls -- basically moving most of `suite/*` to the gravity repo).

Consider:

Gravity X (5.5? 8.0?) tests depends on robotest Y.  Robotest Y depends on Gravity Z (7.0).  How do we keep  X and Z sane?  It is possible, but would take further design than a naive "robotest as used today depends on gravity".

When these values are redeclared we have the possibility to have `case` statements to appropriately branch based on gravity feature detection.

3) I'd like to avoid the situation where we need to maintain & port to a separate branch of robotest for each active release in gravity.

I know this is where planet & satellite have ended up.  However, because Gravity is test code: if we get to the point where each version of gravity needs a custom version of robotest, I'd prefer to move the gravity version specific parts of robotest into the gravity repo, where they'll be managed more like unit tests.  There may still remain some core (e.g. provisioning) that lives on as a library in this repo.  This will allow for the tight dependencies you're suggesting, without the overhead of "a patch for gravity" -> "a patch for robotest" -> "a patch for gravity" whenever gravity breaks robotest's expectations.

Let me know if you have questions about this.

infra/gravity/cluster_status.go
wadells: Why would running multiple status in parallel take longer timeout than running a single status?  What is a reasonable offset in your opinion?

infra/gravity/node_commands.go
a-palchikov: Ok, this definitely sounds like trouble, so it's been cleared.

infra/gravity/cluster_status.go
a-palchikov: Ah, yes, I did not think about them running in parallel when asking the question - sorry, it's fine.

infra/gravity/cluster_status.go
wadells: Updated to your suggestion in 0809da1. In addition to strong typing, this code is also way less complex than `utils.Collect` which is a benefit in its own right.

I'll update the testing in a sec once my run finishes.

I'm still interested to hear back if you think `utils.Collect` has value at all (other than offering backend logic for `utils.CollectErrors`).

wadells:
Testing updated with an install that includes 0809da1.

